PR #1, opened 2026-08-29
Trace format v0, project scaffolding and conformance tooling

yi-here:
## What this changes

Establishes the artifact other people build against — the normative trace format — plus the licensing, governance and CI that decide whether anyone is willing to.

Reviewed separately from the implementation on purpose: the CLI and the viewer are replaceable, the format is not.

## Why

A format nobody else can implement is a file extension. Three things make v0 a real format:

- **`spec/orca-trace-v0.md` is normative**, and the JSON Schema is its machine-readable form. TypeScript types are derived from runtime constants that `schema-parity` tests prove equal to the schema enums — so the format cannot quietly become TypeScript-shaped.
- **A conformance runner** (`scripts/conformance.mjs`) validates every shipped trace against the schema in CI. Any implementation in any language can run the equivalent check.
- **CC BY 4.0 on the spec**, so other implementations may quote it freely. Apache-2.0 on the code, for the explicit patent grant that matters when third parties ship plugins.

The neutrality check (`scripts/check-neutrality.mjs`) fails CI if a vendor plugin imports anything beyond the published `@orcareplay/plugin-api`. Enforcing the OrcaRouter boundary mechanically rather than by good intentions is the point — nobody writes an adapter for a project that might grow a privileged path for its owner.

`examples/traces/run_9f2c14a03b71` is checked in doing four jobs at once: documentation, contributor fixture, conformance input, and something to evaluate the project with before installing it. Its README answers three real questions about the run using nothing but `jq`, which is the actual test of whether the format is greppable.

## Tests

Written before the implementation throughout.

`packages/schema` — schema-parity tests plus validator coverage. Writing the payload test first surfaced a real bug: an overlapping `oneOf` branch was silently accepting corrupt blob digests as opaque inline data. The schema now requires that any object carrying `$blob` be a well-formed blob reference, and the spec says so normatively.

`packages/plugin-api` — compile-time contract tests, so the published interface shape cannot change without failing the build.

## Checklist

- [x] `npm run check` passes locally
- [x] Tests were written before the implementation
- [x] Spec and schema updated together
- [x] No new runtime dependencies (`ajv` + `ajv-formats` only, for schema validation)
- [x] Commits signed off

---
_Generated by [Claude Code](https://claude.ai/code/session_01C2yYKwaBo9qTzYVVHyCcrY)_


yi-here:
## CI red here is environmental, and this PR is the clearest proof

All four checks fail 1–3 seconds after being queued with **zero steps executed**, and job logs return HTTP 404 because nothing was produced. That is a job failing at runner assignment, before checkout.

This PR is the useful data point: **its diff is only Markdown and a JSON Schema.** There is no build here to fail. It fails identically to #2, which strongly indicates GitHub Actions is not yet enabled or entitled for this repository — worth checking **Settings → Actions → General** and that runner minutes are available.

Verified locally from a clean `node_modules`, running exactly what the workflow runs:

```console
$ npm ci && npm run fmt:check && npx tsc --build && npx vitest run
  All matched files use Prettier code style!
  627 passed (627)
$ node scripts/conformance.mjs      # 27 events checked, 0 failure(s)
$ node scripts/check-neutrality.mjs # 0 failure(s)
```

Full detail, including a real lockfile bug the clean-install check surfaced, is in [#2's comment](https://github.com/Continuum-AI-Corp/OrcaReplay/pull/2#issuecomment-5461876718).

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #2, opened 2026-08-29
Implementation: record, replay and fork, end to end

yi-here:
## What this changes

The working system behind the three commands. Stacked on #1 — review that first.

Recording a Claude Code run, replaying it offline, then forking it onto two models

That animation is one real session, captured with `NO_COLOR=1`; `docs/media/transcript.txt` is the raw output, so every line in it can be traced back to a command that actually produced it.

## Why

Model APIs are stateless, so an agent resends the whole conversation every turn — including the previous turn&#39;s tool results. A proxy in front of the model therefore sees the entire loop, which is why **nothing here patches the agent**: it gets a couple of environment variables and is otherwise untouched.

Exact, fork and compare are not three subsystems. They are one proxy with a cursor: the position in the recorded stream where it stops serving from disk and starts serving from the network.

**Packages:** `core` (append-only writer, reader, content-addressed blobs, checkpoint derivation, redactor) · `fs-capture` (shadow git index) · `providers` (Anthropic + OpenAI translation both ways, pricing) · `proxy` (matching ladder, three cursors, opt-in TLS interception) · `adapters` (claude-code, codex, opencode, generic) · `shell-shim` · `mcp-shim` · `viewer` (one self-contained HTML file) · `cli` · a read-only Python SDK.

## Tests

**1067 TypeScript tests across 60 files, plus 180 Python.** Written test-first throughout, including an end-to-end suite that drives a real child process, a real proxy and a real trace on disk.

Verified against the **globally installed** binary rather than `dist/`: clean clone → `npm ci` → `npm run build` → `npm install -g ./packages/cli`, then `doctor` (10/10, both shims), `record`, `list`, `show`, `checkpoints`, `replay`, `fork`, `compare`, `export`, `scrub --dry-run`, `gc --dry-run` — ten steps, all exit 0. Seven error paths return exit 1 with an actionable next step. The exported HTML has zero external references and zero console errors.

### Bugs the tests caught before they shipped

- **A hostile `.gitignore` with `!.env` re-included secrets.** Git&#39;s `info/exclude` is *lower* precedence than the workspace&#39;s own ignore file, so the specified exclusion mechanism was defeatable. Now enforced with pathspecs no ignore file can override — and `GIT_LITERAL_PATHSPECS` in the ambient environment silently disarmed those too, so the git environment is scrubbed.
- **Shannon entropy alone flags ordinary identifiers.** `getUserAuthenticationTokenFromRequestHeaders` scores 4.08 bits/char. Since agent traces are mostly source code, the rule as specified corrupted exactly the payloads the trace exists to preserve. A mixed-alphabet guard fixes it with essentially no loss of recall; the spec is updated to match.
- **OpenAI&#39;s `prompt_tokens` includes cached tokens and Anthropic&#39;s `input_tokens` does not**, so every cached OpenAI turn was double-billed.
- **Fork restored from a fresh shadow store** rather than the source run&#39;s, so the tree object was missing — precisely the silent-wrong-state failure checkpoints exist to prevent.
- **Embedded git repositories** were stored as gitlinks whose contents never entered the shadow store, so a fork would have debugged a workspace quietly missing a subtree.

### Bugs only running it as a user could find

The suite was green at 851 tests while several of these were live. Each was found by using the built binary against a real agent, or by opening the viewer and looking at it.

- **Replay matched nothing on any real recording.** The writer spills a payload over 4 KB as `JSON.stringify(payload)`, and every wire body is a string — so a spilled body was stored as a quoted, escaped JSON string literal while the same body under the limit was stored as itself. Invisible in tests because every fixture sat under 4 KB; fatal in practice because a real harness spills on turn one.
- **The halt was a 409**, which sits in the Anthropic and OpenAI SDKs&#39; default retry sets, so the harness re-sent the unmatched request until its budget ran out and then stalled. The operator saw a hang, not a reason.
- **Rung 2 served recorded answers to different questions.** Its tolerance had a 64-character floor — half the body on a short request.
- **Exact replay ignored the filesystem.** A harness reads files into the conversation *and* writes absolute paths into its tool calls, so replaying in place re-read what the recording changed, and replaying a copy elsewhere got a permission refusal.
- **The proxy buffered instead of streaming**, so every turn of an interactive session appeared to hang for its full duration.
- **Cross-provider compare could not have worked.** `--model gpt-5.2` on an Anthropic-recorded run sent an Anthropic body to api.anthropic.com.
- **The viewer&#39;s timeline could not scroll.** Grid items default to `min-height: auto`, so every event past the fold was unreachable.
- **`causes` was empty on every tool result ever recorded.** A flat event list where the format promises a chain.
- **A failed tool call rendered as `ok`.** The recorder writes `is_error`; the viewer checked `error` and `ok`, neither of which it sets.
- **`orca gc` would have deleted a user&#39;s project directory.** The worktree guard keyed on `parent_run` plus a `$TMPDIR/orca-*` path shape; replay traces then acquired `parent_run` with the user&#39;s own cwd. Caught by the feature that introduced the hazard, and re-keyed on `fork_point`.

Two invariants have tests whose failure messages state them outright: replay makes **no network call at all**, and **anything below rung 1 must carry a divergence**.

## Since the first review round

- **`orca setup` / `orca models`.** Stores a gateway, key and default model list at `~/.config/orca/config.json` (0600 in a 0700 directory), and asks the gateway what it actually serves so a wrong URL is an answer now rather than a 401 mid-comparison.
- **Exact replay writes its own trace**, satisfying spec §4 on the primary path — a run with `parent_run` and no invented `fork_point`, holding divergences and unmatched requests. `last` skips replay traces, so the commands that default to it still mean the recording.
- **Opt-in TLS interception**, for a harness that ignores base-URL variables. No new dependency: a ~200-line encoder-only DER writer covering exactly the ASN.1 RFC 5280 needs. The CA is per-run, `0600`, expires within a day, is trusted only through the child&#39;s own environment, and is deleted at exit including on a crash. A mutation-verified test proves the key cannot reach a recording.

## Since the second review round

Every item found by running the built CLI as a user would, or by writing the test that had been missing. All mutation-verified.

- **MCP capture had never worked, and `orca doctor` said it did.** `resolveShimEntry()` asked for a subpath the package&#39;s `exports` map blocks, so the fallback ran every time and resolved to the *library*. Every stdio MCP server was rewritten to launch a module that exports and exits — `--mcp-config` silently **broke the agent&#39;s MCP servers** as well as recording nothing. Behind it, the shim writes `{name, dir}` and the recorder read `{server, direction}`.
- **`--tls-intercept` was accepted by every command and honoured by one.** `replay --model`, `fork` and `compare` all launch a live agent and all three threw the flag away.
- **A fork could not be forked.** A checkpoint is derived from a filesystem snapshot (spec §3), and forks recorded none — so `orca compare last` straight after a fork failed with &#34;this run has no checkpoints&#34;. Branching had a depth limit of one.
- **`orca scrub` could destroy the evidence it was cleaning** (plain `writeFile` truncates first), and had no `--dry-run` — the one command whose effect cannot be undone was the only one you could not look at first.
- **An unsealed run was reported as tampered.** Same `ok: false` for a killed recorder and an edited file.
- **Exact replay leaked a copy of your working tree** into `$TMPDIR` per invocation, in a directory `gc` deliberately will not sweep.
- **Three copies of the secret-header set had drifted** — `api-key` (Azure) and `x-goog-api-key` (Google) were known to the TLS interceptor alone.
- **Two tests that could not fail**, including `expect(mono_us).toBeGreaterThanOrEqual(0)`.

## Since the third review round

- **MCP capture stopped at the fork point, and replay could not start the agent.** Replay and fork launch the same agent the recording did, so they shared the blind spot record had before `--mcp-config` existed. The recording knew which config to use and had nowhere to write it down (`manifest.argv` holds the agent&#39;s arguments; `--mcp-config` is orca&#39;s), so record now leaves an `mcp_instrumented` note and replay reads it back.
- **`route.decision` was declared and emitted by nothing.** Spec §2 defines it as &#34;a gateway chose a model&#34;, and on a fork orca *is* the gateway — it substitutes the model, picks the wire format and picks the origin, and was doing all three silently.
- **The `Provider` layer is deleted** — ~400 lines of production code and 570 of tests, all seventeen exports checked individually for consumers outside the package (zero). The interface stays published in `@orcareplay/plugin-api`.
- **The shipped example trace recorded a `checkpoint` event**, which spec §3 forbids. Removed, seq renumbered dense with `causes` remapped, goldens recomputed from the Python and TypeScript implementations independently and cross-checked.
- **OrcaRouter is the default gateway.** Visible in the prompt, one Enter to accept, `--gateway` overrides. A default, not a redirect: with nothing configured, `orca record` still proxies the agent&#39;s own calls to whatever provider it was already talking to.
- **`npm ci` printed 5 vulnerabilities, 1 critical** — all dev toolchain. vitest 2.1 → 4.1 and esbuild 0.24 → 0.28: **0 vulnerabilities**, suite unchanged.
- **README rewritten for a first-time reader**, plus two Mermaid diagrams, a comparison table, and a release workflow gated on a computed topological publish order.

## Since the fourth review round — pointed at a real agent for the first time

An earlier version of this description said the system had been &#34;verified repeatedly against the real Claude Code binary&#34;. That was true of `record` and of the CLI surface, and **not** true of `replay` or `fork`: what had been exercised end to end was a fake agent against fixtures. Recording an actual Claude Code session fixing an actual bug — a money-splitting function whose parts did not sum to the total — and then replaying it broke four things at once. Each is a class of bug no fixture can contain, and each is fixed and mutation-verified here.

- **A sixteen-character drift scored 217,568.** Structural distance was the longest common prefix and suffix of the whole serialized request. Claude Code carries a session id in its system prompt *and* another inside a tool description, so the ~200 KB lying between the two counted as changed: rung 2 could not fire on any real recording and every request fell to rung 3 or halted. Distance is now summed over the leaves that actually differ, and per line within a leaf — the same failure existed one level down, where four `duration_ms` floats scattered through fifty lines of test output scored 1,289 instead of the real ~40.
- **Redaction made rung 1 structurally unreachable, and nobody noticed because no fixture holds a secret.** The recorded side holds `<secret:kind:hash8>`; the incoming side holds the value, and the digest is salted per run by design so it could not be reproduced even in principle. The matcher now redacts the incoming request with the same policy and compares the *kind* of secret rather than the digest — reported as a `minor` divergence, since folding is an approximation. Spec §4 gains both rules, because a reimplementation hits the same wall.
- **Redaction broke every fork, which is the feature this is pitched on.** A `tool_use` id is 25 characters of mixed-case base62, so the entropy sweep took it; replay hands recorded bytes back to the agent, the agent echoes the id on its next turn, and the API answers `400 messages.1.content.2.tool_use.id: String should match pattern '^[a-zA-Z0-9_-]+$'`. Anthropic `thinking` blocks fail the same way on their `signature`. Protocol values that must round-trip are now exempt from the entropy *guess* — never from the pattern rules, so a credential parked under a key called `id` is still redacted by shape, with a test holding that line. Worth flagging for review: the first version of this fix **passed its own tests and was still broken**, because a response body is stored as a string inside the event JSON and the scanner sees `\"id\":\"toolu_…\"`; the tests now use the stored form.
- **A replayed agent re-runs its own tools.** Orca does not intercept tool execution — that is the price of not patching the harness — so the agent really runs `npm test` again and it really reprints its own durations. A request whose differences are *all* inside tool output is now served from the recording as a `major` divergence instead of halting the replay. Deliberately narrow: a changed question, or a check that went from failing to passing (`is_error` is not blanked), still refuses at rung 4.
- **`replay.done` said `matched=1 total=13` for a replay that reused all thirteen**, because rung 1 is unreachable whenever anything was redacted. It now reports `reused=7/7 exact=2`.

That recording now replays offline end to end — `reused=7/7 exact=2 divergences=5 unmatched=0 exit=0`, every approximation named in the trace — and a fork of it from a checkpoint reaches the same tree the recording did, having solved the bug on its own. The README says all of this, including the last bullet, which is a permanent property rather than a bug that went away.

## Also since the fourth round

- **The README is translated into seven languages** — Japanese, Simplified Chinese, Korean, German, French, Spanish and Arabic, in `docs/i18n/`, with a language row on each. Full translations, not summaries: a check asserts each file has the same heading, table, code-fence, image and diagram counts as the English original, that every Mermaid block is structurally identical to the English one known to render, and that all 200 relative links across the eight files resolve. Terminal output stays verbatim because it is output. Arabic is wrapped in a single `dir="rtl"` block.
- **`scripts/publish-order.mjs`** computes the topological publish order and fails on a cycle, a non-exact internal dependency, or a version disagreement; `.github/workflows/release.yml` is tag-gated and gates on it. All eleven packages were packed and installed from tarballs into a clean project to prove it.

## Known gaps

- **`docs/orcarouter.svg` is a placeholder wordmark, not the real brand asset.** The sandbox this was built in blocks `orcarouter.ai` (403 at the CONNECT tunnel), so the logo could not be fetched. Drop the real file at that path; the README needs no edit.
- **`api.orcarouter.ai` was never probed live**, for the same reason. `orca setup` degrades correctly (config saved, clear warning) if the probe path is wrong.
- **TLS interception has not been proven against a harness that genuinely ignores base-URL variables** — the case it was built for. Covered by tests against real HTTPS origins on both the record and fork paths, but Claude Code honours the base URL.
- **A fork runs in a scratch worktree**, so a harness that embeds its working directory in the prompt diverges below the fork point and goes live earlier than asked. Reported rather than hidden, but it costs tokens; forking in place would mean a live agent editing your files, which is not a default anyone should get by accident.
- **`checkpoint` is the one declared event type nothing emits**, deliberately: spec §3 derives it at read time.
- **The shipped example trace is hand-written** and is meant to be — a fixture our own writer produced could not show that someone else&#39;s implementation is possible.

## CI

Every check on this PR is red for an environmental reason, not a code one. Across ~50 runs on five branches — including #1, whose diff is Markdown and JSON Schema, and #4, which is someone else&#39;s — every job fails 2–5 seconds after queueing with `created_at == started_at`, zero steps executed and 404 logs. It has now reproduced identically on fifteen consecutive head SHAs. Nothing runs. It needs **Settings → Actions → General** and org Actions billing.

One check count did change on purpose: the OrcaCode Review job now skips instead of failing when `ORCAROUTER_API_KEY` is absent, taking the red checks from six to five.

Because CI has never executed, everything above was validated locally: `fmt:check`, `tsc --build`, `vitest run` (**1067 tests across 60 files**), `scripts/conformance.mjs`, `scripts/check-neutrality.mjs`, `scripts/publish-order.mjs`, `pytest python/` (**180**), and `npm audit` (**0 vulnerabilities**).

## Checklist

- [x] `npm run check` passes locally
- [x] Tests were written before the implementation
- [x] Spec and schema updated together (the entropy guard; §4 on redaction folding and per-field distance)
- [x] Commits signed off


yi-here:
## CI is red for a reason outside this PR — needs a repo setting, not a code change

All four checks (`check (node 20)`, `check (node 22)`, `trace format conformance`, `plugin API neutrality`) fail **1–3 seconds after being queued, with zero steps executed**. Job logs return HTTP 404 because nothing was ever produced. That is the signature of a job failing at runner assignment, before checkout.

Evidence that it is not this diff:

- It reproduces identically on #1, whose diff is **only Markdown and a JSON Schema**. That cannot fail a build.
- Four runs across three different commits, all failing the same way in the same few seconds.
- No step ever ran, so no code in this PR was executed.

The likely cause is GitHub Actions not being enabled or entitled for this repository yet (new repo, or org runner/billing policy). I cannot change that from here — it needs someone with repo or org settings access to check **Settings → Actions → General**, and that Actions has runner minutes available.

### What I verified instead

The exact command sequence CI runs, executed locally from a **clean `node_modules`**:

```console
$ rm -rf node_modules && npm ci        # 67 packages
$ npm run fmt:check                    # All matched files use Prettier code style!
$ npx tsc --build                      # clean
$ npx vitest run                       # 627 passed (627)
$ node scripts/conformance.mjs         # 27 events checked, 0 failure(s)
$ node scripts/check-neutrality.mjs    # 0 failure(s)
```

### One real CI bug this did surface, now fixed

Running `npm ci` rather than trusting a warm `node_modules` found that **`package-lock.json` was out of sync** — the CLI package had been renamed to `orcareplay` after the last lockfile regeneration, so `npm ci` refused to install. Every job would have died at the install step even with working runners. Fixed in `335e458`.

The same clean-install pass also surfaced two genuine races, both fixed in that commit: trace writes were firing in parallel and colliding on git's `index.lock`, and snapshotting a live workspace could catch the agent mid-write (`short read while indexing`). The end-to-end suite went from roughly one failure in three to 10 passes out of 10.

I am not re-running the checks again — three runs across three commits is enough to establish the pattern, and a fourth will fail the same way until the repository setting changes.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #23, opened 2026-09-03
Add an Agent Skill for reading recorded runs

xizhuomengcontin:
## Orca-Code-Review — push 3

|---|---|---|
| P0 | 0 | 0 |
| P1 | 0 | 0 |
| P2 | 0 | 0 |
| P3 | 0 | 0 |

✅ no blocking findings

Adds `skills/orca-replay/SKILL.md`. One file, nothing else touched.

## What it is

An [Agent Skill](https://agentskills.io) — the open format Claude Code, Cursor, Gemini CLI, OpenClaw, Hermes, goose, OpenHands and others all read. A folder with a `SKILL.md`: the `name` and `description` load at agent startup (~100 tokens), the body loads only when a task matches.

## Why this is not a second copy of the MCP server

`orca mcp` already gives an agent the verbs. What it cannot give is the judgement about when to reach for them. Today a user asks "why did you delete my migration file" and the model answers from its own context, because nothing tells it that this is a question with a recorded answer. Six tool descriptions are not enough to change that; a skill description loaded at startup is.

So: MCP is the capability, the skill is the instruction layer on top. They ship together.

## What it tells the agent

The rule is *when a question is about something that already happened, read the trace before answering*. Beyond that it pins down three things this repo cares about and an agent will otherwise get wrong:

- **`recorded` vs `inferred` must survive into the answer.** "The trace shows the `rm` at step 14 removed it" and "this looks like the `rm` at step 14, going by timing" are different claims. Flattening them into one confident sentence is exactly what `orca_graph` labels edges to prevent, and the skill says so, including naming the rule when leaning on an inferred edge.
- **Replay before explaining.** It is free and offline, so there is no reason to commit to an explanation without confirming the failure is deterministic. It also spells out that `reused=3/5` on an interactive recording is a complete replay of what the user asked, not a partial one.
- **`orca_compare` spends real money.** Ask first, say how many models and forks that means, and do not run it to satisfy curiosity the user did not express.

It also covers `worktree: true` when someone is watching their editor, the exact/approximate replay split for argv-vs-typed prompts, and `orca scrub` before sharing a trace.

## Validation

```console
$ skills-ref validate skills/orca-replay
Valid skill: skills/orca-replay
```

`skills-ref` is the reference implementation from [agentskills/agentskills](https://github.com/agentskills/agentskills). 118 lines, ~1,300 tokens of body (spec recommends under 5,000 and under 500 lines), `description` 501 chars (limit 1,024), `compatibility` 159 (limit 500), `name` matches the directory.

One snag worth knowing if you run the validator on Windows: `skills-ref` crashes with a `UnicodeDecodeError` because `validator.py` calls `read_text()` without an encoding and picks up the system GBK codepage. `PYTHONUTF8=1` works around it. Upstream bug, not this file.

## Distribution

The point of the format is portability, and it opens a channel that does not gate on stars or downloads:

- **[ClawHub](https://clawhub.ai)** — OpenClaw's official skill registry, self-serve publish. Dry run passes (`slug: orca-replay`, free; `version 1.0.0`). Publishing there releases that copy under MIT-0 per ClawHub's terms — the file here stays Apache-2.0.
- The skill lists — ComposioHQ/awesome-claude-skills (74k), sickn33/agentic-awesome-skills (46k), VoltAgent's two (52k/34k), travisvn (15k). Several of those want a skill with some adoption behind it, so those are better submitted later than immediately.

Note that agentskills.io itself is the specification and a client showcase, not a skill registry — nothing to submit there.

### DIFF
--- a/skills/orca-replay/SKILL.md
+++ b/skills/orca-replay/SKILL.md
@@ -80,9 +80,16 @@ conversation prefix, so the model is the only variable. Grade with `verify` — 
 exit code is the verdict, e.g. `"npm test"` or `"npx tsc --noEmit"`. Pick the fork point with
 `orca_checkpoints` and pass it as `from`.
 
-**`orca_compare` spends real money.** Every model named is actually called. Ask before running it,
-say roughly how many models and forks that means, and do not use it to satisfy your own curiosity
-about a question the user did not ask.
+**`orca_compare` uploads the recording to other people's models, and spends real money doing it.**
+Each model named receives the same files and conversation prefix the original run had — so whatever
+that run touched (source, prompts, configuration, anything a credential was pasted into) is sent to
+every provider behind those model ids.
+
+Before calling it, tell the user *what* will be sent and *to whom*, not only how many models and
+roughly what it costs. Approving a bill is not approving a disclosure, and the two need separate
+answers when the recording is from a private codebase. `orca scrub` exists for the cases where the
+comparison is worth running but the trace is not safe to send as-is. Never run it to satisfy
+curiosity the user did not express.
 
 ## If there is no recording yet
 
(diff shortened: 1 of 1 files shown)


xizhuomengcontin:
The red `check` on node 20 and 22 is not from this PR.

`npm run fmt:check` fails on three files this branch does not touch:

```
[warn] capture/capture.mjs
[warn] capture/render-run.mjs
[warn] capture/render.mjs
```

The same three fail on `origin/main` — I checked them out separately with this repo's `.prettierrc.json` and ran Prettier against them, and main's own CI has been red since #19 and #20. `.prettierignore` also lists `*.md`, so the file this PR adds is not part of the check at all; it passes Prettier anyway.

`plugin API neutrality`, `python sdk` and `trace format conformance` all pass.

Fix is `npx prettier --write capture/*.mjs`, but that belongs in its own PR rather than mixed in here. Say the word and I will send it.

PR #39, opened 2026-09-07
proxy: match a dialect on the path, not on the path plus its query string

xizhuomengcontin:
## Orca-Code-Review — push 1

| Severity | Count |
|---|---|
| P0 | 0 |
| P1 | 0 |
| P2 | 0 |
| P3 | 0 |

✅ no blocking findings

One line, and the same file already does it forty-five lines below.

`onDecrypted` — the recording side — passed the raw request path to `selectDialect`.
`onInterceptedRequest` — the replay side — has always done `request.path.split('?')[0]` first. So
the same proxy matched a dialect when replaying a request and missed it when recording one.

## What it costs

Measured against one controlled origin, same body, only the query differing:

| path | recorded as |
|---|---|
| `/v1/chat/completions` | `model.request` + `model.response` |
| `/v1/chat/completions?api-version=2026-02-01` | `net.request` + `net.response` |

The second is not a corner case: **Azure OpenAI's endpoints carry `?api-version=` and it is not
optional there.** A run against one records traffic that cannot be replayed and cannot be forked,
and the only thing the run says about it is

```
warn tls.unclaimed_path path=/v1/chat/completions?api-version=2026-02-01
  detail="decrypted and recorded, but no wire dialect claims this path"
  next="docs/plugins.md to add a dialect for it"
```

— pointing the operator at the docs to write a dialect that already exists and already matches.

## The change

A dialect matches an endpoint, and a query string is not part of the endpoint. The path stays in
the trace in full, because that is what the agent asked for; only the match ignores the query. The
test asserts both halves.

## How it was found

By driving oversized responses at a controlled origin for an unrelated check, and noticing that
even the small control case came back as `net.*`. The test is written the way that would have
caught it: the same request as the neighbouring `records an intercepted model API call as a model
exchange` test, with a query string on the endpoint.

Mutation-checked: restoring the raw path turns exactly that test red and nothing else. Full suite
compared per-test against `main` — same failures either way; the one difference,
`e2e.test.ts > does not leave a copy of the working tree in the temp directory`, is a temp-dir race
under parallel load that passes 3/3 in isolation with this change and 2/2 without it.

### DIFF
--- a/packages/proxy/src/server.ts
+++ b/packages/proxy/src/server.ts
@@ -389,7 +389,12 @@ export async function createProxy(options: ProxyOptions): Promise<ProxyHandle> {
    * falls through to `net.request` / `net.response`, where it is described but not interpreted.
    */
   function onDecrypted(exchange: NetExchange): void {
-    const dialect = selectDialect(dialects, exchange.path);
+    // On the path alone. A dialect matches an endpoint, and a query string is not part of the
+    // endpoint -- Azure OpenAI's are all `?api-version=...`, so matching the raw path recorded
+    // those runs as opaque network traffic that cannot be replayed or forked, while telling the
+    // operator no dialect claimed the path. `onInterceptedRequest` below has always split it off
+    // for exactly this reason; the recording side had not.
+    const dialect = selectDialect(dialects, exchange.path.split('?')[0] ?? exchange.path);
     if (dialect && exchange.method === 'POST' && !exchange.requestTruncated) {
       try {
         // Codex's HTTPS fallback currently labels this SSE body as application/json. The wire
--- a/packages/proxy/test/tls-intercept.test.ts
+++ b/packages/proxy/test/tls-intercept.test.ts
@@ -664,6 +664,30 @@ describe('TLS interception', () => {
     expect(netExchanges).toHaveLength(0);
   });
 
+  it('recognises a model API call whose endpoint carries a query string', async () => {
+    const handle = await startProxy([`127.0.0.1:${model.port}`]);
+
+    await through({
+      proxyPort: handle.port,
+      host: '127.0.0.1',
+      port: model.port,
+      trust: [runCa.certPem],
+      method: 'POST',
+      // What Azure OpenAI's endpoints look like, and they are not optional there.
+      path: '/v1/chat/completions?api-version=2026-02-01',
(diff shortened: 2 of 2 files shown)


xizhuomengcontin:
### These two conflict in `server.ts`, and here is the resolution, tested

#33 and #39 both touch `onDecrypted`, two lines apart, so whichever lands second hits a conflict.
Confirmed rather than assumed — `git merge-tree` reports it in both directions:

```
CONFLICT (content): Merge conflict in packages/proxy/src/server.ts
Auto-merging packages/proxy/test/tls-intercept.test.ts        ← this one is clean
```

The test file merges cleanly; only `server.ts` needs a hand.

**Both sides are additive and both are wanted.** #39 changes what is passed to `selectDialect`;
#33 changes the condition below it. Neither supersedes the other:

```ts
    // On the path alone. A dialect matches an endpoint, and a query string is not part of the
    // endpoint -- Azure OpenAI's are all `?api-version=...`, so matching the raw path recorded
    // those runs as opaque network traffic that cannot be replayed or forked, while telling the
    // operator no dialect claimed the path. `onInterceptedRequest` below has always split it off
    // for exactly this reason; the recording side had not.
    const dialect = selectDialect(dialects, exchange.path.split('?')[0] ?? exchange.path);
    // `status === 0` means no response header was ever seen: the client abandoned the call before
    // the origin answered. The request is still worth keeping -- it is what the agent asked --
    // but a model exchange with no response is not one, and inserting it in the replay set gives
    // the matcher an entry that can answer nothing and cannot be forked. It goes below as network
    // traffic instead, which is what it is.
    if (
      dialect &&
      exchange.method === 'POST' &&
      !exchange.requestTruncated &&
      exchange.status !== 0
    ) {
```

I merged the two locally and applied exactly that, then checked it rather than eyeballing it:

- Both branches' new tests pass together — the query-string case, the abandoned-exchange cases,
  the capture-limit case and the `model.response` causes case, 63 of 64 in those three files.
- Full suite compared per-test against `main`: no new failures. The one difference is
  `e2e.test.ts > does not leave a copy of the working tree in the temp directory`, a temp-dir race
  under parallel load that passes 3/3 in isolation on either branch.

The local merge was a check only — it is not pushed, and neither branch has been rebased onto the
other, so both still review on their own.

PR #53, opened 2026-09-08
fix: capture OpenCode APIs from the official model catalog

askdkc:
## Orca-Code-Review — push 2

|---|---|---|
| P0 | 0 | 0 |
| P1 | 0 | 0 |
| P2 | 0 | 0 |
| P3 | 0 | 0 |

✅ no blocking findings

## What this changes

Derive monitored API bases from OpenCode's [official JSON catalog](https://models.opencode.ai/api.json), including model-specific endpoints. Refresh at startup with a bundled offline fallback, preserve original destinations, and skip catalog refresh during replay.

## Why

`orca record opencode` missed providers outside its small hardcoded allowlist. OpenCode supports many AI providers and keeps adding more; using its official catalog lets capture coverage follow those additions without maintaining a separate provider list.

## Tests

- All 196 bundled API templates and installed-OpenCode capture tests pass.
- Formatting and typechecking pass. With isolated settings/temp paths, the full suite reports 1,925 passed, one npm-registry timeout, and two Node 26 file-handle cleanup errors.

## Checklist

- [ ] `npm run check` passes locally (failures noted above)
- [ ] Tests were written before implementation (regression tests added afterward)
- [x] Trace format unchanged
- [x] Commits signed off


xizhuomengcontin:
Thanks for this — I reproduced the bug you're fixing and it's real. On `main`, a POST to `/anthropic/v1/messages` or `/coding/v1/messages` matches no dialect, so an OpenCode run against a catalog provider gets recorded as opaque network traffic that can't be replayed or forked:

```
main:  /anthropic/v1/messages  -> exchanges=0  dialect=undefined
main:  /coding/v1/messages     -> exchanges=0  dialect=undefined
```

Your change fixes both. But one line in it is wider than the fix needs, and the extra width causes a worse problem than the one being fixed.

## The problem

```ts
matches: (p) => p.startsWith('/v1/messages') || p.endsWith('/messages'),
```

`endsWith('/messages')` claims **any** path ending in `/messages` for the Anthropic dialect, and `anthropicDialect` is second in the list (`codex, anthropic, openai, responses`), so it gets first refusal. Because orca picked the origin on this route, `upstreamPath = target.requestPath` — so the request is rewritten to `/v1/messages` **and sent to `api.anthropic.com`**, with the original path discarded.

Reproduced through the real CLI, with two stub upstreams so it's visible which one receives the call. The agent posts an **OpenAI-shaped** body to `/api/v2/chat/messages`:

```
$ orca record generic-openai \
    --upstream-openai    http://127.0.0.1:39601 \
    --upstream-anthropic http://127.0.0.1:39602 \
    -- node agent.mjs        # posts {model:"gpt-5.2", messages:[...]} to /api/v2/chat/messages

main:      (neither stub hit — stays NET traffic, path preserved)
this PR:   ANTHROPIC-UPSTREAM POST /v1/messages
```

Deterministic — 3/3 identical runs at the CLI level, and 3/3 at the unit level.

And the recording that comes out is not just mislabelled, it's falsified:

```
model.request   {"model":"gpt-5.2", "dialect":"anthropic", "path":"/api/v2/chat/messages"}
model.response  {"model":"claude-opus-5", "stop_reason":"end_turn", ...}
```

`orca show` renders that as `MODEL gpt-5.2` followed by `MODEL claude-opus-5` — a turn the run never had. A non-model endpoint also gets claimed: a Slack-shaped `{text, channel}` body posted to `/v3/conversations/messages` came out as an Anthropic model exchange too, so the `toCanonicalRequest` try/catch isn't a backstop here.

This also reaches the TLS-intercept path, where `onDecrypted` calls the same `selectDialect` — there orca doesn't reroute, but any intercepted host's POST to a `/messages` path gets recorded under the wrong dialect, which is what replay and fork read to pick a wire format.

## Suggested fix

One word narrower:

```diff
- matches: (p) => p.startsWith('/v1/messages') || p.endsWith('/messages'),
+ matches: (p) => p.startsWith('/v1/messages') || p.endsWith('/v1/messages'),
```

That still covers every case your comment names — `/anthropic/v1/messages`, `/coding/v1/messages` — and I checked it against your own suite:

| | `endsWith('/messages')` | `endsWith('/v1/messages')` |
|---|---|---|
| `/anthropic/v1/messages` | fixed | fixed |
| `/coding/v1/messages` | fixed | fixed |
| `/api/v2/chat/messages` (OpenAI body) | rerouted to Anthropic | untouched |
| `/v3/conversations/messages` (non-model) | rerouted to Anthropic | untouched |
| this PR's tests | pass | **pass** |

Nothing in the PR depends on the wider form, so the width is pure downside. If some catalog provider does mount Anthropic at a bare `/messages`, that'd be worth handling explicitly rather than by a suffix that also catches everything else.

Worth adding a test that pins the boundary, too — a path ending `/messages` that is *not* Anthropic must not be claimed by the Anthropic dialect. Otherwise the next person widening this has nothing stopping them.

## Everything else looked good

Tested against latest `main` (merges clean, no conflicts):

- full suite 3x: **1934 passed / 23 failed**, identical each run, failure set byte-for-byte the same as `main`'s baseline (the 23 are pre-existing Windows-only failures — POSIX modes, SIGTERM, symlinks). +214 tests from this PR.
- `orca record -> replay -> show -> events -> export -> checkpoints -> graph -> list` 3x, 21/21 checks each
- `conformance` 57/0, `neutrality` 0, `fidelity --check` 0 regressions
- `test/integrations/run.mjs` (the framework suite): **7 checked, 0 failed** — worth calling out since this PR touches `dialects.ts`

Two details I thought were nicely handled: embedding `refreshOpenCodeApiBases` into the generated plugin via `toString()` keeps the catalog fetch in the agent's process, so non-OpenCode recordings pay nothing for it; and setting `OPENCODE_DISABLE_MODELS_FETCH=1` in `replayExact` keeps replay offline. The fetch itself is careful too — 3s timeout, `credentials: 'omit'`, `redirect: 'error'`, and no URL in the fallback warning.

Happy to re-run all of the above once the matcher is narrowed.

askdkc:
Fixed in `6011e21`. 

You're right; the suffix match was too broad.

It now uses `endsWith('/v1/messages')`, preserving `/anthropic/v1/messages` and `/coding/v1/messages` support.

Added regressions for both unrelated paths through the proxy and TLS interception. All four reproduced the incorrect routing or classification before the fix and now pass.

PR #56, opened 2026-09-09
fix(cli): do not detect an agent when the command line already named one

xizhuomengcontin:
## Orca-Code-Review — push 1

| Severity | Count |
|---|---|
| P0 | 0 |
| P1 | 0 |
| P2 | 0 |
| P3 | 0 |

✅ no blocking findings

## The bug

In a LangGraph project, on a machine with Claude Code installed:

```console
$ orca record -- python my_graph.py
info adapter.detected id=claude-code
info recording run=... adapter=claude-code
API Error: 400 unknown provider for model claude-opus-5
```

Claude Code ran. `python my_graph.py` was handed to it as arguments and never launched. The run recorded, so the failure reads like a bad API key rather than like orca having started the wrong program — which is what makes it worth fixing rather than documenting.

## Why it happens

Detection is machine-wide by construction. Every adapter's `detect` is the same call:

```ts
// packages/adapters/src/detect.ts
export async function detectAgent(binaries: string[], homePaths: string[]): Promise<boolean> {
  if (homePaths.some(homeDirHas)) return true;
  for (const bin of binaries) if (await hasBinary(bin)) return true;
  return false;
}
```

`claude-code` is `detectAgent(['claude'], ['.claude'])` — is `claude` on PATH, does `~/.claude` exist. `cwd` never enters into it. That is the right question for "is this agent installed at all", and the wrong one for "what should I launch". A command after `--` was being answered with it.

## The fix

A command after `--` is not a hint about the environment. It is the answer.

## Why `generic-openai` and not `exec`

Both decline to guess what the command is; only one of them redirects anything. `exec` was the first choice here, and it was measured:

```
warn tls.intercept_required adapter=exec ...
base_url: https://api.openai.com/v1/
```

Nothing recorded. `exec` points the agent nowhere — correct for a Go binary with its origin compiled in, and wrong as a default, because most things run this way read a base-URL variable and are captured by the three `generic-openai` sets. Sending everyone to `--tls-intercept` to record a Python script would be a worse answer than the bug it replaces.

## Verification

Four tests, all end to end against the built CLI, all mutation-tested — reverting the branch turns two red, and restoring `exec` turns the fallback test red for its stated reason.

The third test supplies its own `PATH` and `$HOME` with a `claude` shim on them. Without that it launched whatever real agent the machine had, which is both non-deterministic and how it first showed up as an `EBUSY` under parallel load. Removing the shim now turns that test red, so the environment control is itself pinned.

| | |
|---|---|
| `record-detection.test.ts` | 4/4, mutation-tested |
| full suite ×2 | 1724 passed, **zero new failures** vs. `main` |
| `test/integrations/run.mjs` ×3 | 8 checked, 0 failed, 0 skipped |

And the original repro, after:

```
info recording run=run_c8f563d6a340 adapter=generic-openai proxy=http://127.0.0.1:60008
GOT: the stub answered
info recorded run=run_c8f563d6a340 events=6 blobs=0 exit=0
```

### DIFF
--- a/packages/cli/src/commands/record.ts
+++ b/packages/cli/src/commands/record.ts
@@ -83,6 +83,22 @@ async function runRecording(
   let adapter: Adapter | undefined;
   if (agentName) {
     adapter = registry.get(agentName);
+  } else if (args.passthrough.length > 0) {
+    // A command after `--` is not a hint about the environment; it is the answer. Detection is
+    // machine-wide by construction — every adapter's is `detectAgent(binaries, homePaths)`, which
+    // asks whether a binary is on PATH or a directory exists under `$HOME`, and neither is a fact
+    // about this directory. So on a machine with Claude Code installed,
+    // `orca record -- python my_graph.py` detected `claude-code` and launched *that*, handing it
+    // `python my_graph.py` as arguments. The run recorded, so the failure looked like a bad key
+    // rather than like orca having started the wrong program.
+    //
+    // `generic-openai` rather than `exec`: both decline to guess what the command is, but only one
+    // of them redirects anything. `exec` points the agent nowhere and warns that interception is
+    // required — correct for a Go binary with its origin compiled in, and wrong as a default,
+    // because most things run this way read a base-URL variable and would have been captured by
+    // setting the three that `generic-openai` sets. Sending everyone to `--tls-intercept` to record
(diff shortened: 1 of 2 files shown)


xizhuomengcontin:
@orcacode-review

xizhuomengcontin:
@orcacode-review

PR #57, opened 2026-09-09
fix(cli): name the route for a command only `--json` answers

xizhuomengcontin:
## Orca-Code-Review — push 1

| Severity | Count |
|---|---|
| P0 | 0 |
| P1 | 0 |
| P2 | 0 |
| P3 | 0 |

✅ no blocking findings

## The bug

```console
$ orca events
error unknown_command
  there is no "events" command
  next: orca help
```

From a tool whose own `--help` had just listed it:

```
  --json         one JSON document on stdout, diagnostics on stderr
                 list · show · events · checkpoints · graph · record · replay · compare
                 · doctor
```

Seven of those eight are terminal commands. `events` is the one that is not, and the failure it gave reads as the CLI contradicting itself rather than as one view being data-only. Exit code 2, so a script sees a usage error for a name the tool documents.

## Why not just add the command

That was the other option, and it is the wrong one. `buildTimeline` is `events.map(...)` — one row per event, nothing folded and nothing dropped — so `orca show` already renders every event in the trace:

```console
$ orca events --json | jq length
6
$ orca show | grep -c '^[0-9]'
6
```

A second table of the same rows would be a worse `show`. What was missing was the sentence saying so, and where to go instead:

```console
$ orca events
error json_only_command
  events answers as data, not as a table
  these are the raw trace records; `orca show` is the same events, rendered
  next: orca events --json
```

## Why a set and not a special case

The bug is a class, not an instance: a future `--json`-only view added without a line here brings the same contradiction back. So `JSON_ONLY` is a set, and a test derives the same set from the two switches in `main.ts` — everything before `jsonMain` is the terminal one, everything after is the `--json` one — and fails when the declaration disagrees with them.

## Verification

| mutation | result |
|---|---|
| `JSON_ONLY = new Set([])` | 2 red |
| add an unregistered `--json`-only case | 1 red, naming the missing name |

| | |
|---|---|
| `json-only.test.ts` | 3/3 |
| full suite ×2 | 1739 passed, **zero new failures** vs. `main` |
| `test/integrations/run.mjs` | 8 checked, 0 failed, 0 skipped |

A genuine typo still answers as a typo — `orca evnets` is unchanged, and there is a test pinning that too.

### DIFF
--- a/packages/cli/src/main.ts
+++ b/packages/cli/src/main.ts
@@ -128,6 +128,16 @@ For a script, an agent, or CI — every command below also answers as data:
 Docs: https://github.com/Continuum-AI-Corp/OrcaReplay
 `;
 
+/**
+ * The commands `--json` answers and the terminal does not.
+ *
+ * A set rather than a special case. It holds one name today, and the point of the shape is that
+ * adding a second `--json`-only view does not quietly bring back `orca events` telling a reader
+ * that a command `--help` had just listed does not exist. `json-only.test.ts` derives the same set
+ * from the two switches in this file and fails when this disagrees with them.
+ */
+const JSON_ONLY = new Set(['events']);
+
 export async function main(argv: string[], cwd = process.cwd()): Promise<number> {
   const args = parseArgs(argv);
   if (args.bool('json')) return jsonMain(args, cwd);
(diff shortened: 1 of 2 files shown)


xizhuomengcontin:
@orcacode-review

xizhuomengcontin:
@orcacode-review

PR #58, opened 2026-09-09
feat(proxy): record which origin answered each model call

xizhuomengcontin:
## Orca-Code-Review — push 7

|---|---|---|
| P0 | 0 | 0 |
| P1 | 0 | 0 |
| P2 | 0 | 0 |
| P3 | 0 | 0 |

✅ no blocking findings

## What was missing

A trace recorded what was sent and what came back, and never who answered it. That is the first question asked of a recording that looks wrong, and it could not be answered from the recording.

`orca record` printed this:

```
info recording run=run_9af4c858f810 adapter=generic-openai proxy=http://127.0.0.1:49310
```

`proxy` is orca's own address — where the agent calls, not where the call goes on to. A gateway left behind in `~/.orca/config.json` redirects every run on the machine, and nothing said so.

The `model.request` event carried `model, dialect, path`; `model.response` carried status and tokens. Neither carried the origin.

## Why per exchange and not in the manifest

There is no single value to put there. Four things decide the origin and three of them are per request:

- an explicit `--upstream-openai` / `--upstream-anthropic`
- the gateway `orca setup` configured
- a `/forward/` base the client announced — which **outranks both** on a live call, because the request names its own destination
- the dialect's vendor default

and a fork that changes provider changes the answer again mid-run. A run can legitimately reach two origins, so the trace has to be able to say two. It is recorded where it is decided.

It goes on `model.response` because it is a fact about the answer, alongside `decoded_from`, which is there for the same reason. Under `--tls-intercept` the origin is the host the agent chose — the case where it is least obvious from the command line and most worth having.

**Absent means an older orca did not record it, never "the vendor's own API."** That is why the vendor default is written down like any other value rather than left implied.

## No schema change

`attrs` is `{"type": "object"}` in `event.schema.json` — free-form. The manifest schema is `additionalProperties: false`, so a field there would have needed a version bump and would have made every new trace unreadable to an older orca. This route costs neither.

## The three surfaces

```
info recording run=... proxy=http://127.0.0.1:49310 upstream=https://api.orcarouter.ai
```

```console
$ orca show
run_8b6c3905cbad  generic-openai@0.2.3  6 events  exit 0
  upstream http://127.0.0.1:60638
```

```console
$ orca events --json | jq '.[] | select(.type=="model.response") | .attrs.upstream'
"http://127.0.0.1:60638"
```

The record line prints only a *configured* upstream. Left alone, each dialect has its own vendor default, there is no single value to print, and `upstream=default` on every ordinary run would train people to skip the line. The trace records the resolved origin either way.

It found something on its first real run, which is the argument for it: `--upstream-openai <stub>` on a machine with a gateway configured prints **both** origins, because `resolveUpstream` falls back per dialect and Anthropic was still pointed at the gateway. That was always true and this is the first time it was visible.

## Verification

| mutation | result |
|---|---|
| drop the field | 6 red |
| read it from the configuration instead of the resolved origin | 2 red — on the forwarded route, exactly where the two differ |
| remove the `orca show` line | 1 red |
| remove the `orca record` field | 1 red |

| | |
|---|---|
| new tests | 12/12 (4 proxy routes, 3 end-to-end surfaces, 5 unit) |
| full suite | 1732 passed, **zero new failures** vs. `main` |
| `test/integrations/run.mjs` | 8 checked, 0 failed, 0 skipped |


xizhuomengcontin:
@orcacode-review

xizhuomengcontin:
@orcacode-review

PR #60, opened 2026-09-09
docs: point the support table at the checks that already back it

xizhuomengcontin:
## Orca-Code-Review — push 2

|---|---|---|
| P0 | 0 | 0 |
| P1 | 0 | 0 |
| P2 | 0 | 0 |
| P3 | 0 | 0 |

✅ no blocking findings

## The gap

`test/integrations/run.mjs` runs eight checks in CI. The README's support table cited **one** of them.

So the evidence existed and the table someone actually reads did not know about it — which matters now, because the next step is documentation PRs into the frameworks' own repos, and the first thing a maintainer there asks is "does this work with our thing, and did you verify it". Two of the frameworks with a passing check were not in the table at all.

## The rows

| row | before | after |
|---|---|---|
| **OpenAI Agents SDK** | `works` | the `AsyncOpenAI` client it is built on, `exact=1`, in CI |
| **Vercel AI SDK** | `works` | an agent posting to a compiled-in origin, `exact=1`, in CI |
| **CrewAI, Aider, OpenHands** | *absent* | the LiteLLM layer they route through, `exact=1`, in CI |
| **browser-use** | *absent* | its own `ChatOpenAI`, `exact=1`, in CI — LLM layer only |

## The wording follows what each check runs, not what it stands for

That distinction is the point rather than pedantry, because these rows exist to be checked by the people who maintain the frameworks they name.

- `litellm` runs `litellm.completion()`. The row says CrewAI **routes through** LiteLLM and that LiteLLM is the layer with an assertion — not that CI drives CrewAI.
- `openai-async` runs `AsyncOpenAI`, the client the Agents SDK builds on. The row says that.
- `browser-use` never opens a browser. The row says **LLM layer only**.

A claim that does not survive being checked is worse than no row.

Each also links the caveat already written down in `docs/integrations.md`: the Agents SDK ships its own traces as a second egress, `LLM(base_url=…)` does not map to LiteLLM's `api_base`, the browser is not driven.

## Verification

A script walks the table and holds every cited claim to the runner:

```
CI 里真实存在的 check: openai-sdk, openai-async, anthropic-sdk, litellm,
                      langgraph-stream, langgraph-tools, browser-use, fetch-hook

✓ every cited check exists in run.mjs
✓ every exact=N matches the exchange count that check asserts
✓ every docs/integrations.md anchor resolves
✓ no row claims CI drives a framework where the check does not

32 passed, 0 failed
```

And the matrix itself, at this commit:

```
8 checked, 0 failed, 0 skipped
```

Documentation only — no source touched. Does not overlap [#59](https://github.com/Continuum-AI-Corp/OrcaReplay/pull/59), which edits lines 3–109; these rows are at 375–385.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@ Connect: [X](https://x.com/OrcaRouter) · [Discord](https://discord.com/invite/Y
 [![Node](https://img.shields.io/badge/node-20%2B-brightgreen)](#install)
 [![Agents](https://img.shields.io/badge/agents-Claude%20Code%20%C2%B7%20Codex%20%C2%B7%20Agents%20SDK%20%C2%B7%20AI%20SDK%20%C2%B7%20any-black)](#which-agents)
 [![Good first issues](https://img.shields.io/badge/good%20first%20issues-12-orange)](docs/good-first-issues.md)
+[![Mentioned in Awesome Claude Code](https://awesome.re/mentioned-badge.svg)](https://github.com/hesreallyhim/awesome-claude-code)
 
 ![Recording a Claude Code run, replaying it offline, then forking it onto two models](docs/demo-cli.gif)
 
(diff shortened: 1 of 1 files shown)


xizhuomengcontin:
Also carries the Awesome Claude Code badge offered in #37, since it is a one-line README change and this PR is already the README one.

Verified before taking it rather than after: the entry is real — [`awesome-claude-code/README.md:554`](https://github.com/hesreallyhim/awesome-claude-code/blob/main/README.md?plain=1#L554) lists OrcaReplay — and `https://awesome.re/mentioned-badge.svg` answers `200 image/svg+xml`. It goes at the end of the existing badge row.

Closes #37.

PR #63, opened 2026-09-10
resolve the upstream before any command creates a run

xizhuomengcontin:
## Orca-Code-Review — push 3

|---|---|---|
| P0 | 0 | 0 |
| P1 | 0 | -2 |
| P2 | 0 | 0 |
| P3 | 0 | 0 |

✅ no blocking findings

Follow-up to #58. The guard added there (`upstreamPlan` throws for an upstream that is not an origin) was resolved *after* the run directory already existed in `record`, `attach` and the fork path — the fork after restoring a checkpoint into a temp worktree as well. #58 fixed the severe member of this family, where `orca replay` had already replaced the working tree; these are the quiet ones.

So a typo in `--upstream-*` left an empty run for `orca list` to show, reading `FROM <parent>@<n>` on a fork as though it had run, plus a restored tree abandoned in `$TMPDIR`.

Measured on the built CLI, before and after:

| command with a refused upstream | before | after |
|---|---|---|
| `orca record --upstream-openai myuser:PASSWORD@gw/v1` | 1 run, 5 files, 0 events | nothing |
| `orca replay last --model gpt-5.2 --upstream-openai …` | 1 run, 0 events + a temp worktree | nothing |

Both still refuse, with the same message.

`upstreamPlan` reads only the arguments and the environment, so there was never a reason for it to run later.

Both halves of the new test were mutated independently — moving either call back down fails it, on its own assertion.

---

## Push 2 — the sibling refusal review found (P1)

Review was right, and the finding reproduces. `--tls-hosts` is refused by `resolveTlsHosts`/`HostPolicy.from`, both inside `setupTlsCapture` — which an exact replay reaches *after* `replayWorkspace` has written the recording over the working tree. Measured on the built CLI at this PR's first commit:

```
$ orca replay last --tls-intercept --tls-hosts '*'
info  replay.restored … note="your files are restored when the replay ends"
error replay.failed
      "*" would intercept every host the agent talks to.

working tree   "export const fixed = false;\n"   ← was "MY UNCOMMITTED WORK"
runs           1 → 2
scratch        orca-safety-5HLeC0 abandoned in $TMPDIR
```

The restore only began at the child's launch, so the throw escaped it, and nothing prints the scratch path — `replay.restored` reports a tree id, not a directory.

**Two halves, because hoisting is only ever as complete as the list of things somebody thought of.**

`planTlsCapture` now runs before `replayWorkspace`, and before the run directory in `record` and in the fork — the same three places the first commit fixed for the upstream, plus the checkout. It also reads `ORCA_TLS_UPSTREAM_CA` now: that one needs the disk, so it had been left out of a function meant to decide from the arguments alone, and being left out is exactly why it went on failing late.

And the release is armed for the whole restored span rather than from the child's launch, with a `released` flag so the ordinary path puts the tree back where it always did. Nothing else in that window is on any list — `createProxy` binds a socket, `RunCa.create` writes a key, `adapter.prepare` runs a harness's own setup.

Measured on the built CLI, before and after. Every row still refuses, with the same message and the same exit code:

| refused invocation | before | after |
|---|---|---|
| `replay last --tls-intercept --tls-hosts '*'` | checkout replaced, 1 run, `orca-safety-*` left | nothing |
| `record … --tls-intercept --tls-hosts '*'` | 1 run | nothing |
| `replay last --from 1 --model … --tls-hosts '*'` | 1 run + a temp worktree | nothing |
| `replay last --tls-intercept`, unreadable upstream CA | checkout replaced, 1 run, `orca-safety-*` left | nothing |

The span guard was measured on its own, by injecting a throw after `openReplayTrace` into the built CLI: without it the checkout holds the recording and the scratch is abandoned; with it the tree comes back and the scratch is removed. Each new test was mutated — moving a hoisted call back below the thing it guards fails it, on its own assertion.

One test helper changed: `grepRunDir` threw ENOENT for a runs root that was never created, which failed the very test asking for the refusal to come first. A directory that is not there holds no files.

### Not fixed here

A throw in that same window **past `createProxy`** leaves the proxy listening, so orca prints the error and then hangs until it is killed. Reproduced with a trace naming an adapter this build does not have. The `catch` that closes the proxy begins at the child's launch, the same way the release did — pre-existing, unchanged by this PR (before: hangs *and* clobbers; after: hangs, tree intact), and worth its own commit.

---

Full suite 2012 passed / 23 failed, the failure set identical to main's own baseline on this machine (Windows: POSIX modes, SIGTERM, symlinks); conformance 57 events 0 failures; fidelity 0 regressions; plugin neutrality 0; integrations 9 checked, 0 failed, 1 skipped.

### DIFF
--- a/packages/cli/src/commands/attach.ts
+++ b/packages/cli/src/commands/attach.ts
@@ -141,6 +141,11 @@ async function runAttached(
   // — so the next command someone types operates on the session that never happened.
   planTlsCapture(args, replaying?.hosts);
 
+  // Before the run directory exists, because this refuses an upstream that is not an origin: resolved after it, a
+  // typo in `--upstream-*` left an empty run behind for `orca list` to show. It depends on nothing
+  // but the arguments and the environment, so there is no reason for it to run any later.
+  const plan = await upstreamPlan(args);
+
   const dir = await ensureRunsDir(cwd);
   const writer = await TraceWriter.create(dir, {
     adapter: { id: adapter.id, version: ORCA_VERSION, harness_version: adapter.harnessVersions },
@@ -154,7 +159,6 @@ async function runAttached(
   let unmatched = 0;
 
   const writes = new SerialQueue();
-  const plan = await upstreamPlan(args);
   const tls = await setupTlsCapture({
     args,
     out,
--- a/packages/cli/src/commands/attach.ts
+++ b/packages/cli/src/commands/attach.ts
@@ -138,8 +138,10 @@ async function runAttached(
 
   // The last thing that can refuse the run, and it runs before the trace exists: a writer created
   // first and abandoned by a throw leaves an unsealed run directory behind, which then wins `last`
-  // — so the next command someone types operates on the session that never happened.
-  planTlsCapture(args, replaying?.hosts);
+  // — so the next command someone types operates on the session that never happened. Awaited
+  // since it grew the one refusal that needs the disk, an `ORCA_TLS_UPSTREAM_CA` that cannot be
+  // read; the guarantee here is unchanged, it just covers one more way to fail.
+  await planTlsCapture(args, replaying?.hosts);
 
   // Before the run directory exists, because this refuses an upstream that is not an origin: resolved after it, a
   // typo in `--upstream-*` left an empty run behind for `orca list` to show. It depends on nothing
--- a/packages/cli/src/commands/record.ts
+++ b/packages/cli/src/commands/record.ts
@@ -20,7 +20,7 @@ import { SerialQueue } from '../serial.js';
 import { appendSnapshot } from '../fs-events.js';
 import type { Output } from '../out.js';
 import type { ParsedArgs } from '../args.js';
-import { persistNetExchange, setupTlsCapture, trustRunCa } from '../tls-capture.js';
+import { persistNetExchange, planTlsCapture, setupTlsCapture, trustRunCa } from '../tls-capture.js';
 import { upstreamPlan } from '../upstream.js';
 import { ORCA_VERSION } from '../version.js';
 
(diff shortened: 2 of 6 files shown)


xizhuomengcontin:
@orcacode-review

xizhuomengcontin:
The P1 reproduces, and is fixed in `bacb639`.

Confirmed against the built CLI at `f26784d` before touching anything: `orca replay last --tls-intercept --tls-hosts '*'` prints `replay.restored`, overwrites the checkout with the recording, then refuses — leaving the operator's uncommitted work in an `orca-safety-*` scratch whose path nothing prints, plus an empty run.

Both halves of the suggested fix, plus the siblings the finding did not name:

- `planTlsCapture` hoisted above `replayWorkspace`, and above the run directory in `record` and in the fork — `record` and the fork leaked an empty run (and, for the fork, a temp worktree) the same way.
- `planTlsCapture` also reads `ORCA_TLS_UPSTREAM_CA` now. It needs the disk, which is why it had been left out of a function meant to decide from the arguments alone — and why it went on failing late, past the restore.
- `release()` made idempotent and armed for the whole restored span, so a throw that never reaches the child's launch still puts the tree back. Measured on its own by injecting a throw after `openReplayTrace` into the built CLI.

All four refusals now leave nothing behind, with the same message and exit code. Three new tests, each mutated. Full suite 2012 passed / 23 failed, failure set identical to main's baseline on this machine; conformance, fidelity, neutrality and integrations all match it too.

One thing the finding's wider suggestion points at that this commit does **not** close: a throw in that window past `createProxy` leaves the proxy listening, so orca prints the error and hangs until killed. Pre-existing and unchanged here — before it hung *and* clobbered, now it hangs with the tree intact. Left for its own commit rather than folded in, since closing it means restructuring the proxy and trace lifetimes and the sealing semantics that go with them.

xizhuomengcontin:
Both P1s reproduce, and both are fixed in `1c3a0ed`.

**Arm the release before the restore.** Right, and the reason is that `replayWorkspace` performs `recorded.restore` itself and only then returns the `Workspace` — so the guard on the next line had nothing to guard with. Fixed inside `replayWorkspace` rather than at the call site, which gives it one postcondition either way: it returns with the tree replaced and a `release` that reverses it, or it throws having already put the tree back. Arming on the following line is sound because of that, not in spite of it.

**Dispose the CA and close the proxy too.** Also right, and the comment claiming otherwise was the worse half. Measured on the built CLI with a trace naming an adapter this build does not have:

```
before:  exit 1, killed by timeout after 20s   (proxy still listening)
         tls/ca.key still in the run directory
after:   exit 1 in ~700ms, no key, and 'error replay.failed' instead of a stack trace
```

`replay` now has `record`'s `abandon`, in `record`'s words — everything from the mint and the bind to the child's exit comes back through it. `replayFork` had the same hole and got the same treatment.

Two things worth flagging from doing it:

- The `return abandon(err)` shape you suggested is what `record` uses, but `record`'s try has no `finally`. Ours does, and it awaits real I/O — so the rejection sat unadopted while the restore ran and Node killed the process with a raw trace before `main`'s catch could print. Needs `return await abandon(err)`. Reduced to eleven lines and confirmed against a caller that does have a `.catch`.
- The last commit's "not fixed here" note about the hang is now stale; this closes it.

Both new tests mutated independently: dropping the restore guard leaves the scratch and fails on that assertion, dropping `mintedCa = tls.ca` leaves `tls/ca.key` and fails on its own. The CA test pins the proxy too, since one left listening keeps the worker alive and the test times out instead.

Full suite 2014 passed / 23 failed, failure set identical to main's baseline on this machine; conformance, fidelity, neutrality, integrations and the Python SDK all match it. The four refused invocations from the previous push still leave nothing behind, and record → replay → fork still runs clean.

PR #66, opened 2026-09-10
publish main to npm, on a channel nobody is opted in to

xizhuomengcontin:
## Orca-Code-Review — push 3

|---|---|---|
| P0 | 0 | 0 |
| P1 | 0 | -1 |
| P2 | 0 | 0 |
| P3 | 0 | 0 |

✅ no blocking findings

npm has been a snapshot of whichever commit someone last tagged by hand. Today `latest` and main's `package.json` **both say 0.2.3**, while main is **15 commits and 5681 lines ahead** of `v0.2.3` — the version number matches and the code does not, which is worse than an obvious lag. What is missing from npm includes #50–#64: `orca quickstart`, the framework checks, the OpenCode catalog capture, and the whole credential-sanitising series.

## Two channels

| | what it is | who gets it |
| --- | --- | --- |
| `latest` | a version someone decided to release | `npm i orcareplay` |
| `next` | main, once a day, as `0.2.4-main.<sha>` | `npm i orcareplay@next` |

`next` is safe for a reason worth stating: every version on it is a **prerelease**, which npm excludes from `npm i orcareplay` even though `0.2.4-main.abc1234` sorts *above* `0.2.3`. `--tag next` never moves `latest`. So main is always installable and nobody is opted in by accident.

The same shape `typescript` uses: `latest` is `7.0.2` while `next` is `7.1.0-dev.20260909.1`, and `npm i typescript` still gets `7.0.2`.

## Daily, not per merge — because a version is not free

npm serves a *packument* (the metadata for every version of a package) on each install, and it grows with the version count:

| package | versions | packument |
| --- | --- | --- |
| `orcareplay` | 7 | 20 KB |
| `typescript` | 3820 | **15.6 MB** |

Roughly 4 KB each, paid by everyone who installs — including people who only ever ask for `latest`. Publishing per merge at this repository's rate (~8 pull requests a day) adds ~2900 versions a year and lands in that second row inside twelve months. Daily is an eighth of it, and it is what `typescript` does for its own `-dev` builds.

**04:00 UTC — noon in Beijing.** GitHub's cron is UTC only and best-effort (commonly minutes late, droppable under load), so it is documented as "some time after noon" rather than as a clock. `workflow_dispatch` is there for when a day is too long to wait.

A schedule has to work out two things, so a `pick` job does:

- **which commit** — main at 04:00 may be one whose checks are still running, or red. It takes the newest commit on main that CI actually passed and checks *that* sha out, with full history, because the newest green commit need not be the tip.
- **whether to publish at all** — on a day when nothing merged, the version it would publish is already on the registry. It asks npm rather than keeping state, so it stays right across a re-run and a hand-run minutes after the schedule.

## `release.yml` gains the other half

A version bump reaching main **is** the release: it publishes `latest` and creates the tag itself, so a hand-pushed tag is no longer the trigger. An ordinary merge still releases nothing, decided by a `decide` job that costs seconds instead of the whole suite. A `v*` tag still works, unchanged.

A release is now a pull request that changes a version number:

```console
node scripts/set-version.mjs 0.3.0 --lock
```

## Why a script for what looks like one edit

Twelve packages name each other by **exact** version, so a bump is **38 fields across 13 manifests** — and `npm version --workspaces` moves each package's own `version` while leaving its dependants naming the previous release. Publish that and `orcareplay@0.3.0` depends on `@orcareplay/core@0.2.0`, which npm will never let anyone replace.

`RELEASING.md` documented the `sed` and the `git checkout package-lock.json` dance this replaces, including *why*: `npm version` writes the lockfile at the one moment when the versions have moved and the pins have not, which is how 0.2.1 shipped with correct manifests and two poisoned lockfile entries. `--lock` fixes the pins first and reconciles the lockfile after, so that window does not exist.

## What I verified rather than assumed

- The `pick` logic run against this repository resolves `90fc463`, derives `0.2.4-main.90fc463`, answers `publish=true`; the same sha a second day derives the same version, which npm already has, so it answers `false`.
- All **12 packages dry-run publish** at `0.2.4-main.90fc463` under `--tag next`, in dependency order, CLI last.
- `set-version.mjs --lock` leaves **zero** `registry.npmjs.org/@orcareplay` entries in the lockfile (the 0.2.1 invariant), and `npm ci` passes on the result.
- The round trip back to `0.2.3` restores every file **byte for byte**.
- **10 new tests**; removing the pin rewriting fails 4 of them, including the `publish-order.mjs` gate.
- `publish-order.mjs` had to be relaxed to accept a prerelease — its check exists to forbid a *range*, and `0.2.4-main.abc1234` is exactly one version. Confirmed by trying it: with the old regex all 12 packages were rejected as "must be an exact version"; a range (`^0.2.4-main.…`) still is. Build metadata (`+…`) stays forbidden, because npm ignores it when comparing versions.
- `vitest.config.ts` now collects `scripts/**/*.test.ts`. Without it the new test would have sat there green and unrun — the failure #62 fixed one layer out.
- Full suite: **2026 passed / 23 failed**, failure set identical to main's baseline on this machine (Windows: POSIX modes, SIGTERM, symlinks). conformance 57 events 0 failures; neutrality 0; fidelity 0 regressions.

Both publishing workflows are `concurrency`-grouped and **queue rather than cancel**: the publish loop walks twelve packages one at a time, and a run cancelled half way leaves some of them on the registry at a version the rest do not name — the one state npm will not let anyone fix.

## Notes before merging

- A `schedule` trigger only takes effect **once it is on the default branch**, so the daily run starts after this merges.
- It reuses the existing `NPM_TOKEN`; no new secret.
- `latest` stays at `0.2.3` until someone bumps a version. This PR deliberately does not, and **nothing has been published** — npm still shows `{"latest":"0.2.3"}` and no `next` tag.

### DIFF
--- a/.github/workflows/publish-next.yml
+++ b/.github/workflows/publish-next.yml
@@ -1,0 +1,118 @@
+# Publish main to npm under the `next` tag, so what is on main is always installable.
+#
+#     npm i orcareplay          # latest — a tagged release, unchanged by this workflow
+#     npm i orcareplay@next     # whatever main is right now
+#
+# Every version published here is a prerelease — `0.2.4-main.<sha>`. npm excludes a prerelease from
+# `npm install <pkg>` even though it sorts above `0.2.3`, so nothing here can reach someone who did
+# not ask for it by name, and `--tag next` never moves `latest`. That is what makes publishing on
+# every merge safe: main is always available, and no user is opted in to it.
+#
+# The version is derived from the commit, never committed back. Bumping a version in main to publish
+# from main would mean a bot with write access to the default branch and a push race between two
+# merges a minute apart; a `<next-patch>-main.<sha>` needs neither, and is unique by construction —
+# which matters because npm never allows re-publishing a version.
+name: Publish next
+
+# After CI, not on the push itself.
+#
+# CI already runs fmt, tsc, the suite on node 20 and 22, conformance, the framework integrations,
+# the Python SDK and plugin neutrality on every push to main. Re-running that here would double it
+# on a repository that merges several pull requests a day, and a second copy of the gate is a second
+# thing to keep in step. `workflow_run` also makes the rule structural rather than remembered:
+# nothing is published unless the checks for that exact commit went green.
+on:
+  workflow_run:
+    workflows: [CI]
+    types: [completed]
+    branches: [main]
+  workflow_dispatch:
+    inputs:
+      dry-run:
+        description: 'Pack and verify without publishing'
+        type: boolean
+        default: true
+
+permissions:
+  contents: read
+  id-token: write # npm provenance: proves the tarball came from this workflow, from this commit
+
+# Queued, never cancelled. Twelve packages are published one at a time in dependency order, so
+# cancelling half way leaves some of them on the registry at a version the rest do not name — and
+# npm will not let that version be re-published to fix it. A burst of merges therefore publishes a
+# prerelease each, in order, rather than racing.
+concurrency:
+  group: publish-next
+  cancel-in-progress: false
+
+jobs:
+  next:
+    # A failed or cancelled CI run must publish nothing. `workflow_run` fires for those too.
+    if: >-
+      github.event_name == 'workflow_dispatch' ||
+      github.event.workflow_run.conclusion == 'success'
+    runs-on: ubuntu-latest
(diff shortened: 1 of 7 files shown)


xizhuomengcontin:
@orcacode-review

PR #71, opened 2026-09-11
record and replay a RAG pipeline

xizhuomengcontin:
## Orca-Code-Review — push 4

|---|---|---|
| P0 | 0 | 0 |
| P1 | 0 | -4 |
| P2 | 0 | 0 |
| P3 | 0 | 0 |

✅ no blocking findings

An index-building run is the first workload orca gets *wrong* rather than misses. Recording IndexRAG and replaying it offline produced a knowledge base of mismatched answers — `unmatched=0`, `exit=0`, and the cache it wrote holding, for each document, some other document's extraction.

Four layers, one per thing a pipeline breaks. Only the third is specific to any project; a second RAG adapter should declare two things and stop.

| Layer | What it answers | Where |
|---|---|---|
| **L0** matching | under concurrency, does each request get *its own* answer? | `proxy/src/matching.ts` |
| **L1** adapter | start the pipeline, take over **two** model origins | `adapters/src/indexrag.ts` |
| **L2** retrieval | replay embeddings and rerank, which no dialect claims | `proxy/src/retrieval.ts` |
| **L3** state | capture the index; reset to the recording's starting point | `fs-capture`, `orca replay` |

## L0 — the answer has to be the right one

A pipeline sends the same *message* many times over: the prompt template is fixed and the document that varies rides in the **system** prompt. The ladder was built for a conversation, where the trailing message is the question — so `sameAsk` was true against every recorded request at once, and rung 2's tolerance (a share of the request's own size) swallowed a whole different paragraph.

Two **ordering** changes rather than another guard:

- an exact match anywhere in the recording outranks an approximate one at the cursor — found by hash, so it costs one lookup and has no window;
- where nothing is exact, the ladder approximates against the **nearest** candidate (lowest rung, then least distance) rather than the nearest to hand.

Rung 3 additionally now requires the history to actually differ, which is what that rung was written for — compaction rewrites `messages`, and it never leaves the count untouched while replacing the system prompt.

A request served from a position other than the cursor is reported as `reordered=N`, **not** as a divergence. Spec §4 reserves divergences for a match below rung 1; calling a worker pool's ordering an approximation made `exact=N/N` unreachable for any concurrent run. What was genuinely never reproduced is still counted by `reused=x/y`.

## L1 — two model origins in one run

Chat goes to a gateway, embeddings to a dedicated provider or a local Ollama. Orca's upstream map is keyed by *wire dialect*, so no `--upstream-*` can express it. A redirected variable that already held an origin now carries it through `/forward/`, so the request announces its own destination instead of being forwarded to the chat origin — or, with a gateway configured and one `--upstream-*` given, to `api.openai.com` with the embedding provider's key on it. `passthroughOrigin` now also prefers the origin configured for the request's own family before any vendor default.

`orca record indexrag` reads `INDEXRAG_EMBEDDING_BASE_URL` from the environment **or** the project's own `.env`, which is where IndexRAG documents it — an adapter that only read the environment would silently record none of that half.

## L2 — embeddings and rerank replay

An embedding is a function of its request, which is a stronger property than a chat completion has. `RetrievalRule` claims `/embeddings`, `/embed` and `/rerank`, keys each call by its own body, and serves the recording back with no dialect and no ladder.

Never forkable, and the type says so rather than a comment: changing the embedding model changes the vector space the index was built in. That is precisely what makes `orca replay <run> --from N --model X` useful — the chat model changes, the index does not.

A batch whose worker pool assembled it in a different order is matched on the *set* and the response rebuilt to the live order. Exact rather than clever: the OpenAI contract carries `data[i].index` and defines `data[i]` as the embedding of `input[i]`. Anything short of an exact permutation halts. Cohere's embed declares no `batch`, because its response has no per-item index to justify one.

Reported on its own axis — `retrieval=N/M` — never folded into `exact`, which is a statement about a ladder these calls do not climb. `--retrieval-store=digest` keeps a sha256 instead of 98 KB of floats per call.

`retrieval.context` derives the query and the retrieved passages from the prompt that carried them, the same reconstruction `tool.call` has always been. Scores and `top-k` are **not** derived, because they are not on the wire and reporting a truncated count as top-k would be a measurement nobody made.

## L3 — the index is the product, and the pipeline resumes

`artifacts.capture` snapshots paths the workspace ignores — `vector_store/`, `cache/`, `dataset/` are IndexRAG's own first three — and cannot reach a credential: the sensitive pathspecs apply to the forced add too, so an adapter declaring `.` still captures no `.env`.

`artifacts.resetBeforeReplay` clears the paths a resuming harness would skip work from, **before** the restore, so they end up holding exactly what they held when recording began — including the populated cache a recording of one mid-pipeline stage started with. Doing it after the restore deleted that stage's own input.

`concurrencyFlag` is declared and applied only by the new `orca replay --serialize`, because clamping concurrency changes the run being replayed.

## Diagnostics that were lying

- `capture.empty` no longer blames a base-URL variable when the proxy's own counter says the agent called it, and says nothing at all about a run whose every call was a replayable retrieval call (`capture.retrieval_only`, an info).
- The unclaimed-path warning now fires on the plaintext route — the one most runs take — and stays quiet where a rule claimed the path.
- A passed-through exchange no longer records `intercepted=true` over a host orca never terminated TLS for.

## Measured

End to end against the real IndexRAG pipeline on `gpt-5.6-sol` with a third-party embedding provider — extraction, bridging, FAISS build and answering as one recorded run, 8 documents at concurrency 10:

```
info replay.done reused=22/22 exact=22 divergences=0 unmatched=0 reordered=18 retrieval=3/3 exit=0
```

offline, in a scratch worktree that had never seen the pipeline, with `cache/mini_faqs.json` and `test_results_*.json` **byte-identical** to the recording's. A fork onto another chat model replays every embedding from the recording and takes only the chat turns live (`replayed=9 live=2 retrieval=3/3`).

Two new integration checks, `rag-index` and `rag-split-origin`; the second asserts *where* each half of the traffic went, which is what distinguishes it from the first.

## Trace format

`retrieval.context` is a new event type, so `SCHEMA_VERSION` moves to **0.2.0** — spec §2.3's own rule for adding one. `net.request`/`net.response` gain `rule`, `replay_key`, `batch_key`, `response_sha256` and `stored`, all purely additive: a reader that does not know them sees the pair it always did.

## Regression

`npm test` on this branch: **14 files failed, 24 tests failed, 2142 passed**. Clean `main` on the same machine: **14 files failed, 24 tests failed, 2036 passed** — the same 14 files and the same 24 tests (Windows path and file-mode failures unrelated to this work). +106 passing. `npm run fmt:check`, `tsc --build`, `scripts/conformance.mjs`, `scripts/fidelity.mjs --check` and `test/integrations/run.mjs` are all green.


ZhenghuaBao:
@orcacode-review

ZhenghuaBao:
@orcacode-review

ZhenghuaBao:
@orcacode-review

xizhuomengcontin:
@orcacode-review

xizhuomengcontin:
@orcacode-review

PR #79, opened 2026-09-14
ci: get main green — a fixture that changes every run, and a workflow that has never run

xizhuomengcontin:
## Orca-Code-Review — push 1

| Severity | Count |
|---|---|
| P0 | 0 |
| P1 | 0 |
| P2 | 0 |
| P3 | 0 |

✅ no blocking findings

`main` has been red since #70 merged, for two unrelated reasons. Neither is
caused by anything currently open.

## `CI / check (node 22)` — a fixture redrawn on every run

`data-uri-redaction.test.ts` built its indexed-PNG palette out of `randomBytes`,
and whether the entropy sweep fires on a given string is a heuristic answer:
`TOKEN` is `[A-Za-z0-9_-]+`, so base64 cuts a credential apart at every `+` and
`/`, and the 14 zero bytes around the 24-byte credential diluted whatever
survived below four bits a character.

Measured, against the built redactor:

```
today's fixture (24 random bytes in a 38-byte palette):  141 misses in 2000 = 7.0%
the whole palette (768 random bytes):                      0 misses in 2000
```

and the file itself, run 40 times, failed twice — in this case and nowhere else.
node 20 passing and node 22 failing was a coincidence of draws, not a version
difference.

So the fixture is fixed bytes now, and it is the threat the paragraph above it
describes at the size it describes: the whole 768-byte palette, 256 entries of
the caller's own, against a one-pixel image referencing one of them. What the
case claims is narrowed to what is true — colour type 3 buys no exemption and
its palette is swept like any other payload, not that every credential small
enough to hide in a mostly-empty palette is found, which no entropy heuristic
promises. `randomBytes` stays everywhere it only has to make an image
incompressible.

20 consecutive runs of the file, no failures. Restoring colour type 3 to
`PNG_CHANNELS` still turns the case red, so the larger fixture has lost none of
its teeth.

This commit is also in #65; whichever lands first, the other merges cleanly.

## `.github/workflows/orca-code-review.yml` — invalid since 2026-08-29

`8c23e5e` put `secrets.ORCAROUTER_API_KEY != ''` in the job's `if`, with a
comment asserting that `secrets` is readable there. It is not — `jobs.<id>.if`
allows `github`, `needs`, `vars` and `inputs` — so the file is invalid, and an
invalid workflow file registers a failed run against the *push* that carries it,
on every branch, with zero jobs and the file's path where its name should be.

Of the last 100 runs of this workflow: 100 `push`, 100 `failure`, zero
`pull_request_target`. It has never run. The commit that did it was called
"stop a check being red for nothing".

Removed rather than repaired, because the reviews are posted by the
`orcacode-review[bot]` GitHub App and have been arriving unaided for the
fortnight the workflow was dead. Repairing it would start a second reviewer on
every PR, spend the paid quota twice, and put a `pull_request_target` job with
that secret back in the way of PR code for a review nobody is missing. The
commit message carries the repaired expression if you would rather have it back
— it is one `git revert` plus that change.

Drop the second commit if you disagree with it; the first stands on its own.


xizhuomengcontin:
Closing: nothing left in it that is wanted.

The flaky-fixture half landed with #65, so merging this now would change exactly one thing — the
63-line deletion of `.github/workflows/orca-code-review.yml` — and that is declined.

For the record, since the file stays: it has been invalid since `8c23e5e` (2026-08-29 18:13:48Z, and
the first zero-job failure is timestamped 18:13:51Z). `jobs.<id>.if` does not accept the `secrets`
context, so the workflow has not run since. The effect is one failed push-event run per branch and
nothing else — it is not a check on any PR, `main` has no branch protection, it bills no Actions
minutes, and the reviews come from the `orcacode-review[bot]` GitHub App, which has been posting
them unaided throughout.

The branch `fix/main-green` is left in place, so the deletion is one click away if the red X ever
becomes annoying. To repair the workflow instead, the expression belongs in the job's `env` — which
may read `secrets` — and the test in the step's `if`, which may read `env`.

PR #83, opened 2026-09-14
fix(cli): say when this layer captured nothing, instead of nothing

xizhuomengcontin:
## Orca-Code-Review — push 2

|---|---|---|
| P0 | 0 | 0 |
| P1 | 0 | 0 |
| P2 | 0 | 0 |
| P3 | 0 | 0 |

✅ no blocking findings

The agent-spans layer has two ways to come back empty, and neither said so.

## The package is not installed

The bootstrap's `except ImportError: return` is silent. Measured — uninstall `orcareplay-openai-agents`, record the same handoff agent again:

| | with the package | **without** |
| --- | --- | --- |
| events | 15 | **11** |
| `agent.start` / `handoff` / `guardrail` | 2 / 1 / 1 | **0 / 0 / 0** |
| stdout | `info recorded … exit=0` | `info recorded … exit=0` |
| warnings | none | **none** |

Two runs, identical output, the whole structural layer off. And it is the **default**: the package is not on PyPI yet (#81 is the path there), so this is what everyone gets today.

Every other capture layer already says when it did not fire — MCP warns `mcp.not_wired`, the proxy warns `capture.empty` and names the variables it set. This one was the exception.

## A record could not be written

`_dropped` was counted from the first version and read by nothing, so the two failures `_write` deliberately swallows — a payload `json.dumps` will not take, and an append that fails — reached the trace as an absence. Swallowing them is right: a debugger must not be able to fail a run. Staying quiet about them is the shape this project keeps finding.

## What changed

The bootstrap writes `{"kind": "unavailable", …}`, `shutdown` writes `{"kind": "dropped", "count": N}`, `agentSpanLosses` reads both off records the drain already has, and `record` warns once per fact. `eventForSpan` ignores both — they are for the operator while the run is on screen, not for a trace that has no event type for "this did not happen".

**`find_spec("agents")` rather than importing it, and gated on the SDK being present at all.** Measured 0.5ms against 1973ms for `import agents`, on an interpreter that starts in 50ms — and this bootstrap is in front of *every* Python process the recording starts. Ungated, `orca record` would warn about a package that `python --version` had no use for.

**Deduplicated on the way out**, because the bootstrap runs in every one of those processes: six lines saying the adapter is missing are one fact, and six warnings read like six failures. Dropped counts add instead, being counts of different records.

## Measured on the way

`sitecustomize` runs while `sys.path[0]` is still the PYTHONPATH entry — the script's own directory is not added until afterwards. So the bootstrap sees PYTHONPATH and site-packages, and nothing the agent happens to sit beside. The end-to-end test shadows both modules through PYTHONPATH for that reason, which also makes it say the same thing on a machine that has the real package installed — every machine running this suite does.

## Tests

7 Python, 12 TypeScript, including a `recordCommand` end to end. Mutation-tested in three places:

| revert | goes red |
| --- | --- |
| `shutdown` back to `return None` | 4 Python tests |
| the bootstrap's `_unavailable(path)` call | the marker test |
| `record`'s `agent_spans.unavailable` warning | the end-to-end test |

Full suite on Windows: **12 failed / 2332 passed**, against a clean-main baseline of **12 failed / 2320 passed** — the same twelve, compared by name. They are the usual Windows set (symlinks, 0600 mode bits, SIGTERM, forks needing a live upstream).


xizhuomengcontin:
The warning this PR adds had a false positive, and it is fixed in `af87744`. Thanks for catching it — it was firing on runs that had lost nothing.

## What was wrong

`find_spec("agents")` is true of **anything** importable called that, and `agents.py` is an ordinary name for an ordinary module. Reproduced with a project whose own `agents.py` holds `ROSTER = ["alice", "bob"]`, `PYTHONPATH=.`, on an interpreter with no OpenAI SDK installed at all:

```
warn agent_spans.unavailable package=orcareplay-openai-agents
  cause="the agent imported the OpenAI Agents SDK, but this package is not installed"
  next="pip install orcareplay-openai-agents"
```

Three things wrong: the SDK was not installed, the agent did not import it, and the reader is sent to install a package they have no use for. The `cause` asserted as fact something the check had never established.

## What it does now

`importlib.metadata.distribution("openai-agents")` — which *distribution* provides the SDK, which is what "the SDK is installed" actually means. Measured on an interpreter that starts in 50ms:

| check | cost | correct? |
| --- | --- | --- |
| `distribution("openai-agents")` | 2.3 ms | ✅ |
| `find_spec("agents")` | 0.5 ms | ❌ any module of that name |
| `find_spec("agents.tracing")` | **2066 ms** | resolving a submodule spec imports the parent — the specific-looking middle option costs what the import it was avoiding costs |

Plus a second gate the bootstrap cannot apply, because it decides before the agent's first statement and can only know what is installed: `record` warns only when the run actually recorded a model exchange. A run with no model call lost no agent structure, and `capture.empty` already reports the larger problem.

And the `cause` now says what was checked: `openai-agents is installed and this package is not`.

## Tests

The lookalike case is asserted at the `-S` unit level, not end to end — on a machine with the SDK installed, which is every machine running this suite, an end-to-end run that warns is warning *correctly* whatever else is on the path, so the distinction cannot be made there. `-S` drops site-packages and keeps PYTHONPATH, which makes all three fixtures a property of what the test writes: SDK installed / a lookalike module / nothing.

Mutation-tested: reverting to `find_spec` turns 3 red including the lookalike; removing the model-exchange gate turns the no-model-call test red.

PR #99, opened 2026-09-18
docs: @ai-sdk/openai does read OPENAI_BASE_URL

### LINKED ISSUE #80: docs: @ai-sdk/openai does read OPENAI_BASE_URL, so the Vercel AI SDK and Mastra sections state the opposite of the measured behaviour
### What the docs say

`docs/integrations.md`, "Vercel AI SDK, and anything with its origin compiled in":

> `@ai-sdk/openai` takes its origin as a constructor argument and **reads nothing from the environment**. The `node` adapter installs a preload through `NODE_OPTIONS` that redirects at `globalThis.fetch` — the one place every JS client agrees on.

The Mastra section inherits the same claim ("the origin is a constructor argument and the environment is not consulted"), and the README support table repeats it for both rows.

### What I measured

`@ai-sdk/openai` **3.0.112** reads `OPENAI_BASE_URL`. Three construction paths, `OPENAI_API_BASE` unset, `OPENAI_BASE_URL=http://127.0.0.1:9/v1`, each probed by letting the request fail and reading the URL it tried:

```text
createOpenAI({apiKey})  -> tried: http://127.0.0.1:9/v1/responses
bare openai provider    -> tried: http://127.0.0.1:9/v1/responses
explicit baseURL arg    -> tried: http://127.0.0.1:11/v1/responses
```

So: the environment is consulted, and an explicit `baseURL` argument wins over it — the ordinary precedence, not "nothing from the environment".

### Why it matters

Three places are affected, and the error runs in the direction of making orca look *more* necessary than it is:

1. **`docs/integrations.md` § Vercel AI SDK** — the premise of the section.
2. **`docs/integrations.md` § Mastra** — "inherits that provider's behaviour exactly" is still true, but the behaviour it inherits is the opposite of what is written.
3. **README support table** — both rows point at `orca record node --` and the fetch preload as the route. If `OPENAI_BASE_URL` works, `orca record generic-openai --` is the simpler route for these two, and the preload is the fallback rather than the requirement.

The `hardcoded_origin.mjs` check is unaffected — an origin compiled into source genuinely does read nothing, and the preload is the right answer there. It is the attribution of that property to `@ai-sdk/openai` that looks wrong.

### Possible explanations, in order of likelihood

- **The SDK changed.** `@ai-sdk/openai` 3.x may have added the env fallback after the check was written. If so the fix is a version note plus updating the route, not a retraction.
- **The check measures something narrower than the doc claims.** `mastra_agent.mjs` and the Vercel check both pass under the preload, which they would whether or not the env var also works — a preload that redirects `fetch` catches the request regardless of how the URL was resolved. So a green check does not discriminate between the two explanations.

Worth adding a check that pins the actual behaviour: set `OPENAI_BASE_URL` with **no** preload and assert the request lands on the stub. That is the assertion that would have caught this, and it would catch a future regression in either direction.

### Environment

`@ai-sdk/openai` 3.0.112, Node 22.23.2, Windows. No preload, no proxy — just the provider and a dead port to read the URL off the error.

cestercian:
## Orca-Code-Review — push 4

|---|---|---|
| P0 | 0 | 0 |
| P1 | 0 | 0 |
| P2 | 0 | 0 |
| P3 | 0 | 0 |

✅ no blocking findings

## Summary

Fixes #80.

The Vercel AI SDK and Mastra sections (and the README support/status lines) said `@ai-sdk/openai` takes its origin only as a constructor argument and reads nothing from the environment. Measured on 3.0.112 that is false since 2.0.41: `createOpenAI({ apiKey })` and the bare `openai` provider honour `OPENAI_BASE_URL`; an explicit `baseURL` still wins.

This update recommends `orca record generic-openai --` for that case, and keeps the `node` fetch preload as the route when the origin is compiled into source.

## Test plan

- [x] Docs-only review of `docs/integrations.md` and `README.md`


ZhenghuaBao:
@orcacode-review

ZhenghuaBao:
@orcacode-review

xizhuomengcontin:
I reproduced this independently before reviewing, and the substance is right — thank you for
picking it up. Three things to change before it lands, one of which is a factual correction to the
new text.

## What I verified

`@ai-sdk/openai` 4.0.71, `OPENAI_API_BASE` unset, three construction paths against a local stub
that records what it was asked for (rather than reading a URL off an error):

```
createOpenAI({apiKey})  -> 127.0.0.1:55478/v1/chat/completions    (env)
bare openai provider    -> 127.0.0.1:55478/v1/chat/completions    (env)
explicit baseURL        -> 127.0.0.1:55479/v1/chat/completions    (explicit wins)
```

And the route this PR now recommends, end to end with a real `@ai-sdk/openai` agent — because a
docs change that sends people down a path nobody walked is worth less than the sentence it
replaces:

```
$ orca record generic-openai --upstream-openai <stub> -- node agent.mjs
record exit 0 | GOT: hello from the stub | model.request: 1  openai /v1/chat/completions

# origin stopped, then:
$ orca replay <run>
replay exit 0 | GOT: hello from the stub
replay.done reused=1/1 exact=1 divergences=0    (the stub saw 0 requests during the replay)
```

So the recommendation holds. Full suite on `main` + this branch: 2469 passed, 0 failed.

## 1. `since 3.0.72` is wrong by a major version — it is `2.0.41`

Measured behaviourally, not by reading a changelog: `OPENAI_BASE_URL` set, provider asked for a
generation, stub records whether it was hit. Both construction paths, each version installed in
turn.

| version | reads `OPENAI_BASE_URL` |
|---|---|
| 1.3.24 | no |
| 2.0.0, 2.0.25, 2.0.37, 2.0.40 | no |
| **2.0.41** | **yes** |
| 2.0.50, 3.0.0, 3.0.20, 3.0.40, 3.0.60, 3.0.71, 3.0.72, 3.0.112, 4.0.71 | yes |

I checked the 3.0 line specifically to rule out the charitable reading — that it worked in 2.x,
broke at 3.0.0 and was fixed at 3.0.72. It did not break: 3.0.0 through 3.0.71 all read it.

As written, the docs tell everyone on 2.0.41+ and on 3.0.0–3.0.71 that they need the `node`
preload, when `generic-openai` has worked for them all along. `since 2.0.41` is the accurate
boundary.

## 2. The check issue #80 asked for is not here

The issue's last section asks for the thing that would stop this recurring:

> Worth adding a check that pins the actual behaviour: set `OPENAI_BASE_URL` with **no** preload
> and assert the request lands on the stub.

That matters more than the prose, because of what the existing checks actually are. Neither
so-called Vercel fixture uses `@ai-sdk/openai`:

- `test/integrations/agents/hardcoded_origin.mjs` — a bare `fetch` to a literal
  `https://api.openai.com/v1/chat/completions`.
- `packages/cli/test/fixtures/vercel-agent.mjs` — a bare `fetch` to a literal `ORIGIN`, described
  in its own header as a "stand-in".

Both are perfectly good tests of *an agent with its origin compiled in*, and the fetch hook is the
right answer for that. But neither can tell us anything about the SDK, which is exactly the gap
the issue identified: a green check does not discriminate between "the SDK ignores the
environment" and "the preload caught it either way". Without a check on the real provider, the
docs and the code can drift apart again and CI will stay green through it.

## 3. The same claim is still in seven places in the source

The PR changes `README.md` and `docs/integrations.md`. The claim it retracts is still asserted,
about `@ai-sdk/openai` by name, in:

| where | what it says |
|---|---|
| `packages/adapters/src/node.ts:14` | "`@ai-sdk/openai` takes its origin as a constructor argument and reads nothing from the environment" |
| `packages/cli/src/commands/record.ts:730` | "`@ai-sdk/openai` is the live example: it takes its base URL only as a constructor argument" |
| `packages/cli/test/vercel-sdk-e2e.test.ts:21,24` | "the only one that could not be reached by setting an environment variable" |
| `packages/cli/test/fixtures/vercel-agent.mjs:5` | "Like `@ai-sdk/openai`, whose default provider takes its origin as a constructor argument and nothing else" |
| `packages/cli/test/fixtures/deaf-agent.mjs:5` | "The Vercel AI SDK's `@ai-sdk/openai` reads no base-URL variable at all" |
| `packages/adapters/test/node-adapter.test.ts:16` | "`@ai-sdk/openai` takes its origin as a constructor argument and nothing else" |
| `test/integrations/agents/hardcoded_origin.mjs:4` | "the Vercel AI SDK takes its origin as a constructor argument and reads nothing from the environment" |

Two of those are load-bearing rationale — `node.ts` is the adapter's reason to exist, and
`record.ts` is the capture-empty warning's. Leaving them makes the docs and the code say opposite
things about the same library, and the next person to read `node.ts` will reintroduce the doc.

**To be clear about what should not change:** the fetch hook and the `node` adapter are still
right, and so is the `capture.empty` warning. Plenty of agents hardcode an origin — that is most
of the point once someone has pinned a gateway in code, as `node.ts` itself says in the sentence
after. It is only the attribution of that property *to `@ai-sdk/openai`* that has to go. The
example needs replacing, not the mechanism.

## One more thing, not a change request

CI has never run on the current head: the latest run is `action_required`, waiting on approval.
The first commit did break `support-claims.test.ts > 'Vercel AI SDK' cites a check that exists and
agrees on the count`, and `142a9bd` fixed it — I confirmed that passes locally on `main` + this
branch — but it would be worth letting CI say so before merging.

cestercian:
thanks for the careful repro — addressed all three:

1. boundary is `since 2.0.41` now
2. added `ai_sdk_openai` integration check (real `@ai-sdk/openai`, OPENAI_BASE_URL, no preload)
3. scrubbed the seven places that blamed `@ai-sdk/openai` for ignoring env

fetch hook / node adapter / capture.empty are unchanged

xizhuomengcontin:
All three are done, and the check is the part worth calling out — thank you.

I pushed one commit to this branch (`ac91198`), which needs saying plainly since it is your branch:
it merges current `main` in, and that is all it does. `main` moved a long way while this sat —
0.4.0 went out, and #93, #109 and #111 landed — so the branch was `CONFLICTING`. The one conflict
is arithmetic rather than a decision: `test/integrations/README.md` opens with the number of
checks, `main` had moved it to twenty-one when #93 added the Haystack pair, and this branch adds
`ai-sdk-openai` on top. Twenty-two, which is what `CHECKS` counts after the merge. Nothing of
yours is touched; if you would rather have done it yourself, say so and I will not do it again.

## What I verified, rather than took on trust

**The version boundary is right.** Installed each version in turn and asked the provider directly
where it sends a request with only `OPENAI_BASE_URL` set: 1.3.24, 2.0.0, 2.0.25, 2.0.37 and 2.0.40
do not read it; **2.0.41** does, and so does every version after it up to 4.0.71. I also checked
the charitable reading — that it worked in 2.x, broke at 3.0.0 and came back at 3.0.72 — and it
does not hold: 3.0.0 through 3.0.71 all read it. So `2.0.41` is the boundary, and it is now what
the docs say.

**The new check discriminates, which is the whole point of it.** Passing is easy; failing for the
right reason is not, so I made it fail:

```
@ai-sdk/openai 4.0.71   ── ai-sdk-openai   1 exchanges, replayed exact with the origin down
@ai-sdk/openai 2.0.40   ── ai-sdk-openai   FAILED
```

and confirmed the mechanism rather than assuming it — under 2.0.40 the request goes to
`https://api.openai.com/v1/responses` with `OPENAI_BASE_URL` pointed at a local stub. So it really
is measuring the SDK's env handling, not something adjacent to it.

The `NODE_OPTIONS` guard fires too. Injecting a preload whose path carries `orca-fetch-hook` gets
`this check is the no-preload path; the fetch hook was injected` rather than a quiet pass, which
is what closes the hole the issue named: a green result here cannot be "the preload caught it
either way". `needsNode` skipping instead of silently passing is the right call as well.

**The attribution is gone from all seven places** — `node.ts`, `record.ts`,
`node-adapter.test.ts`, `deaf-agent.mjs`, `vercel-agent.mjs`, `vercel-sdk-e2e.test.ts`,
`hardcoded_origin.mjs` — and gone the right way. The fetch hook and the `node` adapter keep their
reason to exist; only the example changed, from a library that does read the variable to an origin
compiled into source. That was the part I was most worried would over-correct, and it did not.

## On this tree, Windows

| | |
|---|---|
| whole suite | 2561 passed, 0 failed |
| `prettier --check .`, `tsc --build --force` | clean |
| framework integrations | 22 checked, 1 failed, 4 skipped — the failure is `mcp-stdio`, which fails the same way on `main` and is this machine's Python `mcp` package, not this branch |

CI has never run on this PR's head — the runs sit at `action_required`, which is the approval an
outside contributor's workflow needs. Approving that now; merging once it is green.

xizhuomengcontin:
Correcting my last comment: the commit it mentions (`ac91198`) is gone, and the way I brought this
branch up to date was wrong. Worth saying exactly what happened, since it was your branch.

I merged `main` into it. That resolves the conflict, but the merge base does not move — this
branch forks from `b4df941`, so every one of the nineteen commits `main` had gained since then was
counted as part of this pull request. Twenty-three commits on a three-commit change. That is noise
I put on your PR, and it should have been obvious before I pushed rather than after.

It is undone. The branch now holds your three commits rebased onto current `main`, and nothing
else:

```
92ade5a  docs: @ai-sdk/openai does read OPENAI_BASE_URL
7ee115f  docs: restore exact=1 citation on the Vercel AI SDK support row
c969896  fix: pin OPENAI_BASE_URL at 2.0.41 and drop the false SDK attribution
```

All three still `author=cestercian`. The only edit inside them is the one line the conflict was
about — `test/integrations/README.md` opens with a count, `main` had moved it to twenty-one when
#93 added the Haystack pair, and `ai-sdk-openai` makes twenty-two, which is what `CHECKS` counts.

The tree this produces is byte-identical to the merge I had already tested, so the numbers in my
previous comment still stand for this head: whole suite 2561 passed / 0 failed, `prettier --check`
and `tsc --build --force` clean, framework integrations 22 checked / 1 failed — that one being
`mcp-stdio`, which fails the same way on `main`.

Two force-pushes to your fork in the process, which is more than anyone should have to accept
silently. If you would rather keep your own branch current, say so and I will leave it to you.

PR #102, opened 2026-09-20
match: a regenerated id inside a harness reminder is not a changed question

xizhuomengcontin:
## Orca-Code-Review — push 6

|---|---|---|
| P0 | 0 | 0 |
| P1 | 0 | -1 |
| P2 | 0 | 0 |
| P3 | 0 | 0 |

✅ no blocking findings

MiniMax Code could not be replayed at all. A run that asked the **identical question twice** reached
rung 4 and halted on a recording that had the answer in it:

```
reused=0/1 exact=0 divergences=0 unmatched=8 exit=4
```

## Why

`askDistance` guards the trailing message tightly on purpose — rung 2 is for drift *around* the
ask, and a tolerance wide enough to swallow a different question would serve the recorded answer to
it under a `minor` label. So the budget is 2% of the ask: about **thirteen characters** on a short
one. The ladder expects a regenerated id to sit in the system prompt, which is where every harness
captured so far puts it.

MiniMax Code puts it in the **user's message**. It prepends a `<system-reminder>` block carrying a
session id it regenerates per request and a wall-clock timestamp:

```
<system-reminder>
<agent-context>
  YOUR SESSION ID: mvs_643d348214ea4573bf852652677b7dcf
  date: Sun Sep 20 2026 10:19:41 GMT+0800
</agent-context>
</system-reminder>

Reply with exactly: SAME
```

Measured through this matcher on two recordings of the same question: **34 characters of drift
against a 13-character budget**. Rung 2 refused it; rung 3 wants an identical trailing message and
the trailing message is the part that moved; so it fell to rung 4.

## The tag is not MiniMax's

Five harnesses already in `prompt/` use `<system-reminder>` for injected context — Claude Code,
OpenCode, Kilo Code, MiMo Code, Qwen Code — and they say what it means in their own prompts:

| harness | its own words |
|---|---|
| MiniMax Code | "injected by the harness, **not the user**" |
| Claude Code | "background context, **not user instructions**" |

The ask measurement now takes them at their word and leaves those blocks out of it.

**Out of the ask only.** Rung 2's whole-request distance still counts every character of a
reminder, so a block that grew by a kilobyte is still drift, and a run whose injected context
differs still reports as `minor` rather than exact. What this stops is a regenerated id being read
as a different question.

## What holds the edges

Non-greedy, with three tests written against mutations that survived without them:

| mutation | caught by |
|---|---|
| drop the exclusion entirely | the drift test |
| make the match greedy | two reminders must not swallow the question between them |
| strip everything after an opening tag | a lone opening tag must match nothing |

The greedy mutation **survived the first three tests** — the two-reminder case was added because of
it, not before it.

## Measured after

Three rounds of a real MiniMax Code recording, replayed with egress blocked:

```
reused=1/1 exact=0 divergences=1 unmatched=2 exit=0
```

with the agent printing the recorded answer rather than halting. Full suite: **10 failures before,
10 after — zero new** (all pre-existing Windows-platform failures).

## Not fixed here

`unmatched=2` is a separate matter. MiniMax Code sends two `POST /v1/responses/input_tokens` before
every chat call; no dialect claims that path, and strict replay refuses what it cannot reproduce.

Worth knowing alongside it, and confirmed with a probe rather than inferred: **those two requests
reach `passThrough` in record mode and no `net.*` pair appears in the trace.** A logging origin
shows MiniMax Code sending three POSTs per run; the recording of that run has one `model.request`
and nothing else. So a recording of this harness currently says one request where the agent made
three. That looks like its own bug and I have not touched it.


xizhuomengcontin:
Confirmed and fixed in d47bc35 — the finding is right, and I reproduced it before changing anything.

```
2,000-char reminder, regenerated session id, "fix the auth test" -> "delete the auth test"
  before:  matched=true  rung=2      <- the recorded answer to one question served for another
  after:   matched=false rung=4
```

`askTolerance` now measures `withoutReminders(trailingMessage(recorded))`, so both sides of
`sameAsk` are in the same space.

The existing "still refuses a changed question buried under a reminder" test missed it for the
reason you name, and it is worth recording: its reminder is short, so the budget stayed small
whichever text it was measured on. The new test uses a reminder large enough for the budget to
matter — which is the size the harnesses this change targets actually send. A mutation that reverts
`askTolerance` to the raw ask now fails that test.

MiniMax Code still replays: `reused=1/1 exact=0 divergences=1 exit=0` over two more rounds against
a real recording, because the stripped drift there is genuinely zero.

xizhuomengcontin:
Reproduced, confirmed real, fixed in `7574c94`.

`withoutReminders` deleted every well-formed `<system-reminder>` pair unconditionally, so **when the whole difference lies inside one pair the ask guard has nothing left to measure**. Measured against the example in the review:

| | recorded ask | live ask | result |
|---|---|---|---|
| main | `<system-reminder>fix the auth test</system-reminder>` | `<system-reminder>delete the database</system-reminder>` | rung 4, refused |
| this PR, before | same | same | **rung 2 / minor / dist=19 — the answer recorded for "fix the auth test" served for "delete the database"** |
| after | same | same | rung 4, refused |

The review's point that this needs no harness bug is right: the tag is just text in the body wherever it appears, and a pasted log or a tool result echoing one can be the whole trailing message.

### The fix

Stripping is now conditional — **a reminder comes out only when the text it sits in still says something without it**.

That threshold comes from the recordings rather than a guess. In all sixteen recorded MiniMax Code requests the reminder shares its text block with the question (about 570 characters of injected block against a 30-character ask), so the case this PR exists for still strips and still replays at rung 2:

```
A whole message is a reminder, question changed   before rung2 minor 19  ->  after rung4  OK
B reminder content changed, question outside it   before rung2 minor 19  ->  after rung2 minor 19 (unchanged)
C drifting session id, question byte-identical    before rung2 minor 17  ->  after rung2 minor 17 (unchanged, the point of this PR)
```

One limit, written into the comment: a harness that gave a drifting reminder a content block of its own would get no help, because that block strips to nothing and is counted whole — the same answer `main` gives today. For a measurement whose only job is to refuse, that is the safe direction to be wrong in.

Two tests, both mutation-checked: making the strip unconditional again is caught by the new reminder-only test; never stripping is caught by the three drift tests that were already there. `matching.test.ts` passes 37.

xizhuomengcontin:
The second finding reproduced too, and it is right. Fixed in `8fee348`.

My last round only guarded the case where the reminder **is** the whole message. The review found the narrower and more damaging one: **put a single character outside the block and a question changed inside it disappears again.**

| recorded ask | live ask | main | last round | now |
|---|---|---|---|---|
| `<system-reminder>fix the auth test</system-reminder>\n\nOK` | `…delete the auth test…\n\nOK` | rung 4 | **rung 2 / minor (askDrift 0 vs tolerance 0.08)** | rung 4 |

The review's "the tag is data wherever it appears" is the real point — and it means **there is no position in the message where a reminder's contents can be assumed not to matter.** So the block is no longer deleted anywhere.

### Instead: keep the block and compare it, folding only the tokens a harness regenerates

The list comes from the recordings, not from guessing. Across sixteen recorded MiniMax Code requests the block differs in exactly two places — `YOUR SESSION ID:` and `date:` — and with these rules **the sixteen reduce to one form**.

The id needs both shapes: `mvs_` + 32 hex sits right on the redactor's entropy threshold, so it may arrive raw or as `<secret:high_entropy>` (`foldPlaceholders` has already removed the per-value digest by then), and the two have to fold together. Folding is confined to the block — outside it the same shapes are the user's own content, such as the commit sha they are asking about.

### Drift and budget are deliberately measured on different text

- **Drift** = the whole ask with only volatile tokens folded, so a changed question inside the block still counts.
- **Budget** = a share of what is left after the reminders are removed entirely, so a 2,000-character injected block cannot buy room.

That keeps the first finding fixed as well.

### Five cases

```
A whole message is a reminder, question changed        rung 4  OK
B question changed inside, only "OK" outside (this)    rung 4  OK
C real mcode: id and clock drift, question unchanged   rung 2  OK  (what this PR is for)
D real mcode: id drifts, question changed              rung 4  OK
E a sha outside a reminder is user content             rung 4  OK
```

Four mutations, all caught: deleting the block, folding outside it, taking the budget from the raw ask, and folding nothing at all. `matching.test.ts` passes 39.

xizhuomengcontin:
The third finding reproduced, and it is right. Fixed in `9e57879`. My previous claim that "a question is none of these things" is simply false for a question **about** a sha or a time.

```
<system-reminder>what does commit 4f2a…d4 do?</system-reminder>
<system-reminder>what does commit 9911…99 do?</system-reminder>
-> both fold to the same text, askDrift 0 against tolerance 0.04
-> rung 2, and the answer about one commit came back for the other
```

Same for `what is at 14:30 today?` against `…15:45…`.

### The test should not be shape, but how much else is in the block

A harness reminder is a **body of instructions** with a regenerated field in it; a question that happens to be about a sha is almost entirely sha. Measured:

| | block length | volatile chars | stable text left |
|---|---|---|---|
| the 16 recorded MiniMax Code blocks | 572–578 | 72–78 | **~494** |
| review's counterexample (sha) | 53 | 32 | **21** |
| review's counterexample (clock) | 23 | 5 | **18** |

So folding now requires **at least 64 characters of the block to be something other than a folded token**. 64 is the floor this file already uses, and the two populations are a factor of twenty apart. I started with a ratio instead, and dropped it after measuring: 13% against a 15% threshold is far too thin a margin.

### Two limits, both written into the comment

- A harness whose reminder is shorter than that gets no help. That is the safe direction.
- A **long** block, identical throughout except for a volatile token the user is actually asking about, is still folded. Closing that would mean refusing every long block, which gives up the case this measurement exists for.

### Seven cases

```
A whole message is a reminder, question changed        rung 4  OK
B question changed inside, only "OK" outside            rung 4  OK
C real mcode: id and clock drift, question unchanged    rung 2  OK  (what this PR is for)
D real mcode: question changed                          rung 4  OK
E a sha outside a reminder is user content              rung 4  OK
F this round: a sha inside the block is the question    rung 4  OK
G this round: a clock inside the block is the question  rung 4  OK
```

Three mutations, all caught: folding without the check, the floor at zero, and the floor set so high that the real MiniMax Code drift stops being forgiven. `matching.test.ts` passes 40.

xizhuomengcontin:
The fourth finding reproduced, and it is right. Fixed in `a72bba9`.

My last round's "at least 64 non-volatile characters in the block" **only measured bulk**. Put the same sha question inside a *real* MiniMax Code block and the 494 characters of instructions around it pay for the threshold: the sha folds, and the answer about one commit comes back for the other. The review's sharper version — a 300-character sha question plus 100 characters of scaffolding — goes through the same way.

### Shape can never separate them; the thing that can is a field

In the recorded blocks the volatile values sit in an `<agent-context>` list, one field per line:

```
  agent: Mavis  # display name
  SESSION ROLE: root
  YOUR SESSION ID: <secret:high_entropy>
  date: Sun Sep 20 2026 11:31:14 GMT+0800 (中国标准时间)
```

**One label, one value, the whole value regenerated.** A question is prose: no label, and the token sits inside a sentence rather than being the entire value. So only **the complete value of a labelled line** folds now, and `what does commit 4f2a… do?` is never that, however long the block around it is.

The bulk floor stays as a second guard for the small end the field rule cannot reach: `<system-reminder>\ndate: 14:30\n</system-reminder>` is a labelled field and nothing else, so without a floor a question written that way would fold.

### Three of this repository's own fixtures failed

Not because the rule is too strict, but because **those fixtures are not what the harness sends**: a made-up `Sun 10:19:41` where MiniMax Code sends a full `Date.toString()`, and two reminders with their fields run together on one line. I fixed the fixtures to the recorded shape rather than loosening the rule.

### Nine cases

```
A whole message is a reminder, question changed          rung 4  OK
B question changed inside, only "OK" outside              rung 4  OK
C real block: id and clock drift, question unchanged      rung 2  OK  (what this PR is for)
D real block: question changed                            rung 4  OK
E a sha outside a reminder is user content                rung 4  OK
F short block, the sha is the question                    rung 4  OK
G short block, the clock is the question                  rung 4  OK
H this round: long block, the sha is the question         rung 4  OK
I this round: long block, the clock is the question       rung 4  OK
```

Two of the four mutations survived the first run, and both turned out to be **missing tests** rather than dead code: a labelled value that merely *contains* a sha, and a block that is a labelled field and nothing else. With those written down all four are caught. `matching.test.ts` passes 43.

PR #104, opened 2026-09-20
proxy: an upstream that never answers should leave a mark

xizhuomengcontin:
## Orca-Code-Review — push 5

|---|---|---|
| P0 | 0 | 0 |
| P1 | 0 | -1 |
| P2 | 0 | 0 |
| P3 | 0 | 0 |

✅ no blocking findings

`passThrough` records the exchange **after** the upstream answers. When the upstream never answers,
there was nothing to record: the throw escaped to the server's catch-all at `server.ts:616`, which
hands the agent a 500 — that part was right — and wrote nothing else. No event, no warning, and a
trace saying the run made fewer calls than it did.

## Not an exotic path

`fetchPinned` refuses **every** redirect on purpose: undici strips `authorization` across origins
but not `x-api-key`, so following a `Location` could hand a gateway's key to whoever it names. A
gateway behind a CDN answers an unknown path with exactly that.

Isolated on two real gateways — same MiniMax Code run, only the upstream changed:

| upstream's answer to an unknown path | warning | `net.*` events | total events |
|---|---|--:|--:|
| `404` | `proxy.unclaimed_path` printed | 4 | 18 |
| `301` | none | **0** | **4** |

`api.orcarouter.ai` sits behind a CDN that answers an unknown API path with `301` to its own
marketing site. MiniMax Code posts to `/v1/responses/input_tokens` before every call, so **every
recording of that harness through that gateway lost two requests silently.**

The replay message compounded it — it told the operator that path *"was captured as opaque network
traffic, not as a model exchange"*, about traffic that had never been captured at all.

## The change

The failure is recorded as what it is: `status: 0`, which already means "no response header was
ever seen" for an intercepted exchange, and the reason in the response body. Same run, after:

```
events=8   (was 4)
net.response status=0
  orca did not forward this call: Error: https://api.orcarouter.ai/v1/responses/input_tokens
  answered 301 redirecting to https://www.orcarouter.ai/...
```

The agent still gets its 500. Nothing about the successful path moved.

## The reason is redacted before it is written

`withoutCredentials` handles URLs — userinfo and query strings — and an upstream quoting the call it
refused can quote an `authorization` header that no URL rule touches. §7 says never write it, which
is not the same as writing it and trusting a later layer to catch it. The `Redactor` the matcher
already builds is now shared with this path.

**That test failed first, against the fix as originally written.** It is in the PR because it caught
a real hole, not to decorate one.

## Verification

Three mutations, each caught:

| mutation | caught by |
|---|---|
| don't record the failed call | both tests |
| claim `404` instead of `status: 0` | "records the call it could not forward" |
| skip the redaction | "keeps the credential out of the recorded reason" |

`packages/proxy/test/` — 273 passed, 1 pre-existing Windows permission failure (`0600/0700`).
`passthrough.test.ts` 21/21. The two tests in the full-suite noise that looked related
(`stays quiet about traffic that was never a model call`, `never writes the bank host…`) **pass in
isolation and pass with the whole file** — this machine has around eight flaky tests under the
parallel suite right now, which is what those were.


xizhuomengcontin:
Reproduced, confirmed real, fixed in `8ab0d01`. This is a §7 violation my own commit introduced.

In the same event, `responseBody` was carefully scrubbed while the `host` beside it went in verbatim. Three origins, measured (`createProxy` does not validate origins — `unusableOrigin` is only enforced by the CLI):

```
origin                          host recorded before            after
https://myuser:hunter2@gw bad   "https://myuser:hunter2@gw bad" "https://gw bad"
https://myuser:hunter2@         "https://myuser:hunter2@"       "https://"
gateway.example?key=SECRET123   "gateway.example?key=SECRET123" "gateway.example"
```

`persistNetExchange` files `host` as a `net.request` attr, so this really did reach the trace. The trace's own redactor is no backstop, exactly as the review says: `hunter2` and `key=SECRET123` are both under its 20-character entropy floor and match no shape rule.

### Two problems, not one

Besides the fallback itself, I had **copied the existing `originParts` into a second function** (`netExchangeTarget`) in order to reach it. The copy is gone, the failure path calls `originParts`, and the scrub lives in its catch. `server.ts` is four lines shorter.

That also settles which copy is live: holding a response means undici parsed the URL, so the fallback cannot run on the success path. `passThrough`'s failure path is the only place it does — because the thing that made `doFetch` throw is precisely an origin `new URL` rejects.

The diagnostic value is kept: an operator can still see which upstream failed, just not the password.

Two mutations, both caught: leaving the origin raw, and dropping the host entirely. `passthrough.test.ts` passes 22.

xizhuomengcontin:
Reproduced, and it is right. Fixed in `58c84f9`. This finding goes deeper than what I fixed last round.

`withoutCredentials` is **written for prose** — an error message, a warning line — and pays for that with rules that only fire on text shaped like a URL in a sentence: the scheme-less query rule needs a dotted host, and the userinfo rule cannot cross a space. Six shapes measured, **four go through whole**:

```
gateway.example?key=SECRET123  -> gateway.example      <- the one shape I happened to test
gw?key=SECRET123               -> unchanged
localhost?key=SECRET123        -> unchanged
192.168.0.1?key=SECRET123      -> unchanged            (digits after the dot)
https://user:pa ss@gw          -> unchanged            (space inside userinfo)
https://user:pw@gw bad         -> https://gw bad
```

My test used the only one that works, which is how the gap survived a commit whose entire subject was closing it.

### Cut by position instead of by prose rules

An origin is not prose and does not need those rules. It is **one token**, and everything in it that can hold a secret is positional: query and fragment are whatever follows the first `?` or `#`, userinfo is whatever precedes the last `@` in the authority. Cutting on position works on a string no parser accepted — which is the only kind that reaches here.

`redactString` still runs afterwards, but as a **backstop** for a genuinely high-entropy key rather than the thing doing the work.

### The recorded reason leaked the same way

undici quotes the origin verbatim (`Failed to parse URL from gw?key=SECRET123/v1/...`), so the known origin is now replaced by its scrubbed form before the prose pass ever sees the sentence.

All seven shapes verified, asserting on both `host` and `responseBody`. A valid origin survives unchanged (`http://155.103.159.98:8317/v1`), so the diagnostic value is intact.

Five mutations, all caught: host through the prose rules, host unscrubbed, reason through the prose rules, and dropping either cut from `scrubOrigin`. `passthrough.test.ts` passes 22.

xizhuomengcontin:
Reproduced, and it is right. Fixed in `a91d693`. Three rounds on the same line, and the lesson is the same each time: **a string no parser accepted has no structure to rely on**, so nothing may be kept because of where it sits — only because of what it is made of.

```
upstream = https://gw key=SECRET123   -> new URL rejects it -> the fallback runs
my last version: the token after the space is still "after the last @",
                 so it was kept as the host and written to the trace verbatim
```

### Read the host as a whitelist

Only the characters a host and a port are made of, and the first one that is not ends it. **The path is not returned at all** — the success branch gives `url.hostname`, the exchange already records its request path in its own field, and a path that reached here unparsed could carry a secret as easily as a query could.

`https://user:pass@` has no host in it, so it yields the empty string. That is the honest answer, better than inventing one.

Fourteen shapes verified, including two the review did not name: `gw;key=SECRET123` and an IPv6 authority (`http://[::1]:8080` -> `[::1]:8080`).

### Mutation testing turned up three more problems

Two of the first five mutations survived. They turned out to be **missing tests rather than dead code**:

- drop the query cut, and the `@` in `gw?user@evil.example` is read as userinfo — the trace then names `evil.example` as the upstream that failed
- drop the path cut, and the `@` in `/v1/@scope/pkg` does the same (`withoutCredentials` already names this trap in its own comment)

Writing those two down then exposed a third mistake: my case for the path used `https://host/v1/@scope/pkg key=SECRET123` — the space is in the **path**, which `new URL` percent-encodes rather than rejects, so that origin parses, the fallback never runs, and the case tested nothing at all. With the space moved into the authority, all five mutations are caught.

`passthrough.test.ts` passes 22.

xizhuomengcontin:
Two findings this round: one is real and is fixed in `8eead4d`, the other describes code this PR no longer has.

## The real one: with no `@`, the whole `user:password` becomes the host

`scrubOrigin` strips userinfo by cutting at the last `@` in the authority — and **an origin with no `@` has nothing to cut at**. The whitelist then allowed `:` for ports, so the whole string survived:

```
https://myuser:hunter2  ->  myuser:hunter2
https://user:PASSWORD   ->  user:PASSWORD
```

Both reach this path (`new URL` rejects them, undici throws), and the result is written to the trace as the `net.request` `host` attr and quoted in the recorded reason. `redactString` is no backstop for either — both are under its 20-character entropy floor.

**Userinfo and `host:port` are the same shape until something says which it is.** Digits after the colon are what say it. So a colon is accepted only when a port follows, and `myuser:hunter2` cuts to `myuser`. An IPv6 literal keeps its brackets, which is what separates its colons from that one.

Seventeen shapes verified. The port and IPv6 cases have to survive or the trace stops naming which upstream failed. Three mutations, all caught: a colon without digits, no whitelist at all, and dropping the bracket alternative.

## The stale one

The other finding describes `https://gw key=SECRET123` and the two like it, against the code **before `a91d693`**. Re-checked this round:

```
https://gw key=SECRET123                 -> gw
https://user:pw@host key=SECRET123       -> host
https://user:pw@host:8080 key=SECRET123  -> host:8080
```

`passthrough.test.ts` passes 22.

PR #105, opened 2026-09-20
feat(adapters): record MiniMax Code by moving its data directory

xizhuomengcontin:
## Orca-Code-Review — push 9

|---|---|---|
| P0 | 0 | 0 |
| P1 | 0 | -1 |
| P2 | 0 | 0 |
| P3 | 0 | 0 |

✅ no blocking findings

MiniMax Code (`mcode`) is the first harness here that neither of orca's two capture routes reaches:

- **No base-URL variable.** Its origin lives in `~/.minimax/config.yaml`.
- **No transport route either.** Its HTTP client is Node's `fetch`, which consults neither `HTTP_PROXY` nor `HTTPS_PROXY` — `--tls-intercept` sees nothing and the run ends `capture.empty`.

What it does honour is `MINIMAX_DATA_DIR`. The adapter points that at a directory inside the run and writes a redirected copy of the operator's config there.

## Why relocate rather than rewrite in place

MCode writes to its own config on startup. Measured on a real run: of the config this adapter generated, it put **exactly two lines back** — the built-in `minimax_api` provider's `baseURL` and `apiKey`, restored to its own values. An adapter editing `~/.minimax/config.yaml` in place would be handing the agent its own file to fight over, and a run that is killed never puts it back.

That is also why only **custom providers** can be captured: the built-in managed login's origin cannot be moved by a file. Every capture in `prompt/MCODE/` was taken through a custom provider.

## Two decisions worth reviewing

**Origins go through `/forward/`, so the call still names the host the operator configured.** Pointing them at orca's own default instead would be a captured run talking to a host they never named — the same reasoning `opencode-config.ts` already states.

**Every key becomes the placeholder.** The generated file lands inside the run directory, and §7 says a credential is never written there. Orca supplies the real one for the origin it forwards to (`upstreamHeaders`, what `orca setup` configures). Without it the gateway answers 401 — and the prompt is captured anyway, because it travels in the request.

## Verified end to end

```
orca record mcode -- exec --prompt-mode coding --permission off --max-steps 1 '…'
  adapter=mcode  upstream=http://155.103.159.98:8317   ← the operator's own, carried through
  events=14  exchanges=6
  system prompt: 16,162 chars
  real key anywhere in the run directory: none
```

## `capture: 'config'`

New, for the same reason `transport` exists: "sets no base-URL variable" is otherwise indistinguishable from the broken adapter `redirects-model-traffic` is there to catch. What this adapter redirects cannot be read from the launch environment, and under the contract's synthetic workspace there is no config to copy at all — so the redirect is checked where the file can be read, in `test/mcode.test.ts`. `capture.empty` gets a branch too: telling this operator to add `--tls-intercept` would send them after a route that cannot work.

## The capture profile, and three bugs running it found

`capture/capture.mjs` gains an `mcode` profile with `--prompt-mode` (tui / coding / work — three different prompts from one binary).

- **All three modes mirrored to one filename** and silently overwrote each other, leaving `prompt/MCODE/` holding whichever was captured last under the name of all three. The variant is now part of the name, as `-print` already was for Claude Code.
- **`meta.regenerate` printed a command that reproduced a different capture.** It now carries the mode and `--allow-failed`.
- **A run id and the pwsh path made every capture differ from the last.** Both are scrubbed now, so regenerating produces the same bytes — checked by running it twice and comparing hashes.

The three committed prompts are renamed to what the tool emits and regenerated through it. Each differs from the merged version by **one line**: `activeDataDir`, which is the directory this adapter moves.

## Tests

9 new adapter tests, 4 mutations all caught (not scrubbing the key, rewriting only the first provider, dropping `/forward/`, writing beside the operator's file). Full suite diffed against `main` under the same conditions: **zero new failures** — the six that first appeared were parallel-run flakes and did not reproduce serially.


xizhuomengcontin:
@orcacode-review

xizhuomengcontin:
Reproduced, and it is right. Fixed in `3613009`. This one is worse than I had assumed: I confirmed that `capture.mjs` does `cpSync(dir, traceDest)`, copying the whole run directory into `capture/<model>/trace/`, and the README states plainly that `trace/` is **not** scrubbed. So the key would have left with the capture.

Seven YAML spellings measured against the old regex — **five go through untouched**:

```
apiKey: sk-live-plain                   scrubbed
apiKey: sk-live-commented # rotate me   LEFT INTACT
api_key: sk-live-snake                  LEFT INTACT
api-key: sk-live-dash                   LEFT INTACT
apiKey: "sk-live-quoted"                scrubbed
apiKey: 'sk-live-single'   # note       LEFT INTACT
APIKEY: sk-live-upper                   LEFT INTACT
```

### Now matched by line and split in code

Quoting, a trailing comment and a CRLF line ending are three separate things to get right, and one regex that handled all three would be harder to read than the file it is parsing. A YAML comment is taken to start at ` #`, so a `#` inside a URL is still a fragment. The operator's quoting and comments survive the rewrite.

`hasOrigin` replaces the old `REWRITABLE` test and **reads a line the same way the rewriter does** — otherwise the adapter could write a config that redirects nothing and still report the run as captured.

Re-verified end to end: a real recording against the operator's own gateway, 14 events and 6 exchanges, and neither key from `~/.minimax/config.yaml` appears anywhere in the run directory.

Four mutations, all caught: the old anchor, dropping the `api_key` spellings, folding the comment into the value, and not scrubbing keys at all. `mcode.test.ts` passes 13.

xizhuomengcontin:
Two findings this round. The second is real and is fixed in `c1f2759`; the first does not reproduce, and the measurement is worth putting on the record.

## The real one: `hasOrigin` cannot tell a built-in provider from a custom one

Asking only "is there a `baseURL` line" is true of a **stock install**, where the only ones belong to the built-in providers. The adapter then wrote a redirected config for a run that records nothing — MCode restores those origins over whatever the file says — and left a **14 MB** copy of its data directory inside the trace to show for it.

Origins are now moved only inside `custom_provider:`, so a managed-login-only config takes the untouched branch it was always meant to take:

```
stock (built-in only)  hasOrigin=false  ->  launched untouched, no run-dir copy
mixed                  hasOrigin=true   ->  custom origin redirected,
                                            built-in origin left alone,
                                            BOTH keys -> orca-recorded
```

Keys still come out of **every** section, built-in included. What may be redirected and what may be written down are separate questions — leaving the built-in key in place because its origin is out of scope would be the leak this adapter exists to prevent.

## The other one does not reproduce

The premise is that MCode's startup restore writes **the operator's real built-in credential** back into the generated config. Measured, and it does not:

- Built a config from the real structure with every `apiKey` replaced by a distinct `CANARY-nn-DO-NOT-LEAK`, pointed the adapter at it, and ran a real recording (14 events, 6 exchanges).
- MCode restored the built-in provider's `apiKey` as the literal **`sk-xxx`** — the placeholder from its own bundle — not the canary that was in the operator's config.
- **No canary appears anywhere in the run directory.**

It has nothing else to restore from. Credentials live in `<dataDir>/auth/`, and `MINIMAX_DATA_DIR` moves that too: the run's copy holds `auth/auth.lock` and `auth/prod/cn/mcode-public/auth.lock`, and nothing else. **The relocation is what protects it** — the design choice the finding questions is the reason the leak is not there.

One limit I should state rather than paper over: this account is **not signed in** to the managed login (`~/.minimax/auth/` holds only lock files), so the signed-in variant rests on that mechanism rather than on a measurement of it.

## Also

Two fixtures failed against the stricter gate, and both were configs no machine would have — bare `baseURL:` lines with no section above them. Fixed to the real shape rather than loosening the gate.

Four mutations, all caught: counting any `baseURL`, redirecting built-in origins, skipping keys outside `custom_provider:`, and losing section tracking. `mcode.test.ts` passes 15, and a real recording against the canary config still produces 14 events with nothing of the operator's in the run directory.

xizhuomengcontin:
Three findings this round. The flow-style one is new and real — fixed in `dbb4986`. The other two are the previous round's text repeated against code this PR no longer has; both were addressed in `c1f2759`, and I have re-checked them below.

## The real one: keys survive in shapes the line anchor cannot see

Reproduced exactly as described:

```
custom_provider:
  a:
    options:
      apiKey: sk-live-A                                          -> scrubbed
      baseURL: https://a.example/v1                              -> redirected
  b:
    options: {apiKey: sk-live-FLOW, baseURL: https://b.example/v1}   -> LEFT INTACT
  c:
    options:
      - apiKey: sk-live-LIST                                     -> LEFT INTACT
      - baseURL: https://c.example/v1                            -> not redirected
```

`hasOrigin` was true because of provider `a`, so the file was written with `b` and `c`'s keys in it.

### The fix is not "scrub more shapes"

Keys are now taken out by a pattern that does not care where on the line the field sits, so the flow map and the list item are both scrubbed. But that is not the guarantee. The guarantee is that **a rewrite which cannot prove it did its job does not get written at all**, and three checks make that proof:

- every `api…key:` field carries its value **on the same line** — a value the scrub cannot see is also one a check on the result cannot see;
- no value is a **block scalar**, whose `|` would be replaced while the key stayed on the lines beneath a field that now reads as clean;
- every origin inside `custom_provider:` **points at the proxy after the rewrite**, or a provider orca claimed to redirect would still be talking to its real host, uncaptured and unmentioned.

Failing any of them launches untouched — the same branch a config with no custom provider takes. An uncaptured run is the side to fail on when the alternative is writing a credential down.

```
all block style (ordinary)  -> writes config
flow map                    -> launches untouched
list item                   -> launches untouched
block scalar key            -> launches untouched
value on the next line      -> launches untouched
block and flow together     -> launches untouched   <- the one that shipped
built-in only               -> launches untouched
```

### Mutation testing found a hole neither of us had

`apiKey:` alone, with the value below as an indented scalar, **matched no pattern at all** — so it was neither scrubbed nor noticed, and went into the run directory whole. That is what the same-line check is for. The same run also showed that a post-check on keys could never fire, since it shared a regex with the scrub, so it is gone rather than left in as decoration.

All five mutations are caught: the line anchor, no gate, and each of the three checks removed individually.

## The two repeated findings

**`hasOrigin` cannot tell a built-in provider from a custom one** — fixed in `c1f2759`. Origins are moved only inside `custom_provider:`, so a stock install launches untouched and no run-directory copy is made. Re-checked this round: `built-in only -> launches untouched` in the table above.

**MCode restores the real built-in `apiKey`** — does not reproduce, re-measured this round. Against a config whose every key had been replaced by a distinct `CANARY-nn-DO-NOT-LEAK`, a real recording (14 events, 6 exchanges) restored the built-in provider's `apiKey` as the literal **`sk-xxx`** from MCode's own bundle, and **no canary appears anywhere in the run directory**. It has nothing else to restore from: credentials live in `<dataDir>/auth/`, which moves with `MINIMAX_DATA_DIR`, and the run's copy holds two lock files and nothing else.

The limit I stated last round still stands: this account is not signed in to the managed login, so the signed-in variant rests on that mechanism rather than on a measurement of it.

`mcode.test.ts` passes 17, and a real recording against the canary config still produces 14 events with nothing of the operator's in the run directory.

ZhenghuaBao:
@orcacode-review

xizhuomengcontin:
Five findings this round. Two are new and real — both fixed in `ac03493`, together with a third that running the result found. The other three are the previous rounds' text repeated against code this PR no longer has; re-checked below.

## The two real ones

```
"apiKey": sk-live-…     a quoted name — the quote sits between the name and the colon,
                        so neither ANY_KEY nor KEY_FIELD matched, and the parity check
                        compared 0 against 0 and passed

apiKey: # rotated       the `#` was captured as the "value", so the same-line check
  sk-live-…             passed while the real key sat on the line below
```

Both reproduced: `hasOrigin` true, gate true, and the key in the written config.

## The fix is to stop chasing shapes

This is the fourth round of the same failure, and that is the useful part. Each round added the shape the last one missed — a flow map, a list item, a block scalar, a value on the next line, and now a quoted name and a comment-as-value. That is a blocklist, and a blocklist of YAML spellings is not something anyone finishes.

So it is reversed. **One canonical shape is understood** — optional quotes around the name, a colon, a value that is a single token, optionally quoted, optionally followed by a comment — and a line naming either field any other way is **refused rather than rewritten**. Every shape above now fails the same way, without being enumerated:

```
block style (ordinary)         -> writes config
authMode: api-key present      -> writes config
quoted value                   -> writes config
trailing comment               -> writes config
snake / dash / upper           -> writes config
quoted key name                -> writes config      <- this round
quoted baseURL name            -> writes config
comment mentioning apiKey      -> writes config
built-in origin in flow style  -> writes config
value is only a comment        -> launches untouched <- this round
flow map                       -> launches untouched
list item                      -> launches untouched
block scalar                   -> launches untouched
value on the next line         -> launches untouched
block and flow together        -> launches untouched
built-in only                  -> launches untouched
```

## Running it against a real config caught a third, and that one was mine

Every custom provider in a real `~/.minimax/config.yaml` carries `authMode: api-key`, where `api-key` is the **value**. A scan that looked anywhere on the line read it as a field it could not parse and refused the whole config — a recording that should have captured six exchanges captured **none**. The colon after the name is what tells a name from a value.

Caught only because the end-to-end run came back `events=2` instead of `events=14`. It has its own test now.

Seven mutations, all caught, including the two that only fail on deliberate scoping: keys are scrubbed in **every** section, and a built-in origin's shape is not this adapter's business because it is never rewritten.

## The three repeated findings

**`hasOrigin` cannot tell a built-in provider from a custom one** — fixed in `c1f2759`; `built-in only -> launches untouched` in the table above.

**MCode restores the real built-in `apiKey`** — does not reproduce, measured again this round. Against a config whose every key had been replaced by a distinct `CANARY-nn-DO-NOT-LEAK`, a real recording (14 events, 6 exchanges) restored the built-in provider's `apiKey` as the literal `sk-xxx` from MCode's own bundle, and no canary appears anywhere in the run directory. Credentials live in `<dataDir>/auth/`, which moves with `MINIMAX_DATA_DIR`; the run's copy holds two lock files and nothing else. The limit I stated before still stands: this account is not signed in to the managed login, so the signed-in variant rests on that mechanism rather than on a measurement of it.

**The quoted-key finding was filed twice**, at `mcode.ts:119` and `mcode.ts:236`. Same finding, one fix.

`mcode.test.ts` passes 19; the adapters suite passes 593 with one pre-existing Windows failure unrelated to this PR.

xizhuomengcontin:
Five findings this round. Two are new and real; three are earlier rounds' text against code this PR no longer has. Fixed in `7ef2054`, together with **three more I found by going through the rest of the surface myself rather than waiting to be told** — one of which is worse than either reported finding.

## The two real ones

**Replay could go live.** This is the serious one, and I had not considered the caller at all. `orca replay` calls the same `prepare`, with the operator's *current* environment rather than the recorded one. A config that had lost its custom provider since the recording made the fallback launch MCode untouched — on the real `~/.minimax`, against the real gateway, with the real credential. Nothing reached the proxy, `unmatched` stayed at zero, and the replay reported success.

The adapter cannot tell a record from a replay, so **the data directory now moves whether or not there is a config to put in it**. Measured with an empty directory:

```
mcode exec failed: Sign in to MiniMax to use Agent features. Run `mcode login`, then retry.
```

No call made. The stock-install operator pays for it — `orca record mcode` ends in that message rather than running uncaptured, and MCode fills the directory with its bundled skills either way, about 14 MB — and both are worth it against a replay that silently goes live.

**`apiKey:#rotated`**, with no space before the `#`. The comment strip leaves it in place, so the line parsed as a field whose value was `#rotated` while the real key sat on the next line. A value that begins with `#` is not a value.

## Three more from reviewing the rest myself

- **A CRLF config was refused whole.** The carriage return belongs to the line ending, not to the value; I had dropped that handling in an earlier rewrite. A stricter gate does not throw, so a Windows-written config would simply have captured nothing — silently. Mine, and not reported.
- **An origin `/forward/` cannot carry** — userinfo, a query, a non-HTTP scheme — fell back to orca's own default upstream, sending the run to a host the operator never named. This file's own comment calls that worse than not capturing. Now it refuses.
- Checked the rest instead of assuming: a tab before `#`, a `#` inside a quoted value, a quoted top-level key and a leading BOM all refuse, which is the safe direction.

## Twenty-three shapes

```
block style / CRLF / authMode: api-key / quoted value / trailing comment
snake, dash, upper / quoted key name / quoted baseURL name
comment mentioning apiKey / built-in origin in flow style / port + path origin
                                                        -> writes config

apiKey:#rotated / apiKey: # rotated / tab before #
flow map / list item / block scalar / value on next line / block+flow together
origin with userinfo / with query / ftp:// / built-in only
                                                        -> writes no config
```

Four mutations caught, one per fix. Both ends re-run: a real recording is 14 events with no canary from the operator's config anywhere in the run directory, and the degraded path refuses to start instead of calling out. Full suite diffed against `main`: zero new failures.

## The three repeated findings

**MCode restores the real built-in `apiKey`** — fourth report, does not reproduce. Measured again with per-key canaries: MCode restores the literal `sk-xxx` from its own bundle, and no canary appears in the run directory. Its credentials live in `<dataDir>/auth/`, which moves with `MINIMAX_DATA_DIR`; the run's copy holds two lock files.

**`hasOrigin` cannot tell a built-in provider from a custom one** — fixed in `c1f2759`; `built-in only -> writes no config` above.

**"Refuse a key whose same-line value is a trailing comment"** — describes `ANY_KEY`/`KEY_FIELD`, which `ac03493` replaced. That shape has its own test and refuses; the no-space variant is the one that was still open, and it is fixed here.

`mcode.test.ts` passes 22.

xizhuomengcontin:
Two findings this round, both real and both fixed in `c403e2f` — and both about the same thing: a check that only proves what it can see. The other three are earlier rounds' text; two of those I fixed in `c1f2759` and `7ef2054`, and the third describes `ANY_KEY`/`KEY_FIELD`, which `ac03493` removed.

## `MAVIS_DATA_DIR` survived into the child

This one defeats the isolation I added last commit, and you are right that I built it by naming spellings one at a time. `configPath` reads either spelling; the launch overrode only one. An operator who had set the other kept their real data directory — real config, real origin, real credential — and the isolation bought nothing.

So it is a sweep now, not a list. MCode reads about ninety `MAVIS_*` and `MINIMAX_*` variables, among them **three more data directories** (`MAVIS_RUNTIME_DATA_DIR`, `MAVIS_PARENT_DATA_DIR`) and **four credentials** (`MAVIS_ACCESS_TOKEN`, `MAVIS_PARENT_AUTH_SECRET`, `MINIMAX_API_KEY`, `MINIMAX_CN_API_KEY`). Every one of them in the parent environment is emptied, and the two documented data-dir spellings are then pointed at the run's directory. A variable a later version adds is covered by the sweep rather than by my remembering it.

Verified end to end on the bypass itself — a recording with only `MAVIS_DATA_DIR` set, which the previous commit would have let escape:

```
info recording ... adapter=mcode upstream=http://155.103.159.98:8317
mcode exec failed: ... 401 Invalid API key ...        <- the placeholder, as intended
info recorded ... events=14
canary from the operator's config anywhere in the run directory: none
```

## A credential under a name the rewrite does not know

Right again: the allowlist proves the two fields it understands, and `token:` / `secret:` / `apiToken:` passed every check and was copied out verbatim.

Two nets, because one has a hole I measured rather than assumed. Orca's own redactor is asked whether anything in the result still looks like a secret:

```
sk-live-AbCdEf0123456789XyZ                caught
a real JWT (127 chars)                     caught
40-char random                             caught
AKIA…                                      caught
b8e793df1a6e4b1088eeaa608388afc9  (32 hex)  MISSED
hunter2                                     MISSED
```

The 32-hex miss matters: that string's maximum Shannon entropy is exactly the redactor's threshold, so a hex key never trips it — and hex is the commonest shape an API key takes. So the second net is shape rather than entropy: inside `custom_provider:`, a field this file does not rewrite whose value is **one opaque token of 24 characters or more** is refused. Checked against a real config — names, kinds, model ids and `api-key` all carry punctuation or are shorter, so nothing in it matches.

What survives both is a short low-entropy value under an unknown name, `token: hunter2`. That is in the comment rather than papered over.

`@orcareplay/core` becomes a direct dependency of `@orcareplay/adapters` for the redactor; `proxy` already depends on it and `publish-order` is unchanged.

## Verification

Twenty-eight shapes, four mutations (one per fix, each caught), `mcode.test.ts` passes 24, and the full suite diffed against `main` has zero new failures.

xizhuomengcontin:
Six findings this round, one of them new and real — fixed in `b7fb084`. The other five are earlier rounds' text: `hasOrigin` (fixed in `c1f2759`), `apiKey: # rotated` and `apiKey:#rotated` (both refused, with tests, since `7ef2054`), and the unknown-field-name one (the two nets added in `c403e2f`).

## The real one, and it is a contradiction I wrote myself

The opaque-token net I added last commit was scoped to `custom_provider:`, while the rule three lines above it says keys come out of **every** section — the file lands in the run directory whatever section they are in.

Reproduced:

```
provider:
  minimax_api:
    options:
      authSecret: b8e793df1a6e4b1088eeaa608388afc9    -> written verbatim
```

And the redactor cannot cover it, for the reason the finding gives: a 32-hex string's Shannon entropy is at most the threshold, so the sweep never trips. That value scores **3.58**.

The scan reads the whole file now. Checked for the cost **before** changing it rather than after: a real config has no opaque token under any name in any section, so there is nothing to pay. It also picks up the configs whose section cannot be tracked at all — a quoted top-level key, a leading byte-order mark — where every line reads as the empty section.

```
built-in section, 32-hex     -> untouched
top level, 32-hex            -> untouched
custom section, 32-hex       -> untouched
ordinary built-in block      -> writes config
```

One mutation, caught. `mcode.test.ts` passes 24, the full suite against `main` has zero new failures, and a real recording is still 14 events with no canary from the operator's config in the run directory.

## One thing in the repeated finding worth answering

It also mentions an **origin** under an unrewritten name — `url:`, `endpoint:`. That is not a credential leak, and it is not a capture gap either: MCode reads `baseURL`, so a provider whose origin is written under another name has no origin as far as MCode is concerned and makes no call. The opaque-token net does not catch it because a URL carries punctuation, and I would rather say that than add a rule for a case that cannot fire.

xizhuomengcontin:
Six findings this round. One is new and real — fixed in `6bd85e8`, along with three more of the same kind that came out of going through the rest. The other five are earlier rounds' text against code this PR no longer has.

## The real one: a quoted value walks past the shape test

`authSecret: "b8e793df1a6e4b1088eeaa608388afc9"` — the net tested the shape against the whole captured value, so the leading quote failed it. The case the last commit said it fixed, defeated by two characters.

## Three more of the same kind, and the pattern behind all four

```
authSecret: "b8e7…"                          quoted            <- reported
opts: {authSecret: b8e7…}                    flow map
authSecret:                                  value on the
  b8e7…                                        following line
secrets:
  - b8e7…                                    list item
```

I had written a **detector** the way the **rewrite** is written. That is right for the rewrite — it has to produce correct output, so it must understand a shape before touching it, and refuses what it cannot. For a detector it is backwards: every shape it cannot parse is a value it never looks at.

So it does not parse. It reads the text after the first colon — or the whole line when there is none, which is how a value on the next line is seen at all — and searches for the shape **inside** it. Quotes, braces and commas stop mattering because nothing has to match the whole of anything.

## Three things in a real config that had to be checked, not assumed

Each is now a test:

- `defaultModel: custom_provider:orca-cap/deepseek/deepseek-v4-flash-free` — forty characters of the token alphabet once `/` is in it. That is why `/` is not in it.
- `contextWindowOptionHints:` and `files_api_upload_endpoint:` are both long enough to trip the scan on their own. That is why it starts after the colon.

## I also corrected what the comment claims

"no key may be left behind" was not true, and is now stated as what it is: two nets that catch what they recognise, with a short low-entropy value under an unknown name getting past both. What carries the rest is not detection — `orca push` ships an allowlist of four top-level entries that this directory is not one of, and `capture/<model>/trace/` is documented as the unscrubbed raw run and is the one thing the prompt vault's CONTRIBUTING forbids committing. A value that survives the nets stays on the machine it was already on.

Nine shapes verified, six mutations caught. Two of those mutations found **dead code rather than a bug** — the quote-stripping stopped doing anything once the shape was searched for instead of matched, so it is gone. `mcode.test.ts` passes 24, the full suite against `main` has zero new failures, and a real recording is still 14 events with no canary in the run directory.

PR #111, opened 2026-09-21
feat(adapters): capture ZCode through the provider file it is handed

xizhuomengcontin:
## Orca-Code-Review — push 3

|---|---|---|
| P0 | 0 | 0 |
| P1 | 0 | -1 |
| P2 | 0 | 0 |
| P3 | 0 | 0 |

✅ no blocking findings

## What this changes

Adds a `zcode` adapter, so `orca record zcode` and `orca replay` work on ZCode, Z.ai's coding
agent — plus the `capture.mjs zcode` profile, its harness fixture, and the captured system prompt
as a committed artifact.

## Why

ZCode's provider origin lives in `~/.zcode/v2/provider_config.json`, not an environment variable,
so none of the existing routes reach it: the env redirect has nothing to set, and TLS interception
is the wrong tool for a harness that will happily talk to a plain-HTTP origin if you can tell it
to. `capture: 'config'` is the route, and `ZCODE_PERSONAL_PROVIDER_CONFIG_FILE` is the lever — one
variable, one file, and the operator's own config is read and never written.

This is the second config-route adapter and the first whose file is JSON. That is the whole
difference from MiniMax Code (#105): `JSON.parse` returns a tree and the rewrite walks every node
of it, so "did this reach every key" is not the open question it was for a text rewrite that
closed one unanticipated spelling at a time over eight rounds of review.

Two decisions worth a reviewer's attention:

- **The variable is set even when there is nothing to carry into it.** Replay calls this same
  `prepare` with the operator's *current* environment. An adapter that redirected nothing there
  would let a replay run live on their real providers, spend money, and report success because
  nothing reached the proxy. Same hazard #105 was fixed for.
- **Registered last of the harnesses.** `registry.detect` returns the first adapter that claims
  the workspace, and ZCode's detector reads the machine (`~/.zcode`, or `zcode` on PATH). Any
  earlier slot takes auto-detection away from an adapter that already had it.

A file that will not parse, or an origin `/forward/` cannot carry, gets orca's own single provider
rather than a half-redirect: a partial rewrite would leave whichever provider it missed talking to
its real host, with nothing in the trace to say so.

## Tests

`packages/adapters/test/zcode.test.ts`, 10 tests. The one that earned its place is **"reaches a
key by any name and at any depth"** — six credential spellings, three levels down. It failed on
first run: `isSecret` was `/(^|[-_])(key|token|…)$/i`, which requires a separator before `key`, so
`apiKey` never matched and the operator's real credential would have been copied into the run
directory. Fixed with three boundaries, the camelCase one matched case-sensitively so `monkey` is
not read as a field.

The rest: the adapter contract; registry names; that the operator's file is read and not written;
that the origin is carried through `/forward/` while the key becomes the placeholder; that origins
move wherever they are nested while `apiKeyManagementUrl` is left alone; that a non-JSON file and
an uncarryable origin both fall back to one orca provider; that a provider file is named even with
no config to read; that the fallback asks for a model the catalogue knows; and that user args pass
through.

End to end, twice: `record` → `events=4 blobs=1`, then `replay` → `reused=1/1 exact=1
divergences=0 unmatched=0 egress=blocked`. Re-run after the registration reorder, same result. Two
independent captures produced the same sha256. The operator's `provider_config.json` is unchanged
by hash, and the generated copy holds zero occurrences of a canary key while preserving the
original destination rather than substituting orca's default upstream.

## Checklist

- [x] `npm run check` passes locally — `typecheck` and `vitest` are clean (2554 passed, 0 failed).
      `fmt:check` is green on every tracked file. It does report 130 files, all of them untracked
      local `capture/<run>/` artifacts that are gitignored and never reach CI; three are MCode's
      own bundled HTML skill templates, which Prettier cannot parse.
- [ ] Tests were written before the implementation — not honestly, no. The adapter came first and
      the suite second. The camelCase bug above is what that cost; the test caught it before a
      reviewer had to.
- [x] Spec and schema updated together, if the trace format changed — the trace format did not
      change.
- [x] Commits signed off

## The artifact

`prompt/ZCODE/glm-4-6-system-prompt.md` — 9,084 characters as sent, in one system block, 27 tools,
scrubbed clean. Provenance checked in both directions: its distinctive lines appear in
`zai-org/ZCode` and in the installed runtime, so this is the product's prompt rather than a
wrapper's.

Its `prefix` column in `capture/README.md` is blank because the turn was refused before any tokens
were counted — the adapter carried the operator's own origin through the proxy, that gateway does
not serve `glm-4.6`, and the refusal came back as `200` with an `upstream 400` frame inside the
stream. The prompt is unaffected: it travels in the request, and all 9,084 characters were on disk
before the response arrived.

ZCode's built-in account providers are **not** redirected, and the adapter's comment says so.
`ZCODE_BUILTIN_PROVIDER_CONFIG_FILE` can replace that file, but not with an empty one — ZCode then
refuses to boot at all. Redirecting them would mean locating and rewriting the shipped file, which
is install-specific. A run on an account provider is therefore not captured, the same limit
MiniMax Code has for its built-ins.

Also fixes the `capture.mjs` usage line, which had never listed `hermes`.


xizhuomengcontin:
Both P1s were real. I verified each against the code before changing anything, and found two more of my own while verifying — one of which is in metadata this PR was about to ship.

## P1: keys `isSecret` misses were written into the trace

Confirmed, and the review is right about why. Checked the three regexes against the names it listed: `APIKey`, `APIKEY`, `APISecret`, `accessKeyId`, `jwt`, `auth`, `bearer`, `Authorization`, `Cookie` and `session` all miss. `isOrigin` misses `endpoint`, `host`, `url` and `server`. And the generated config does reach `capture/<model>/trace/` — I confirmed that on the existing capture, where `trace/zcode-config/provider_config.json` is sitting next to the events.

The thing that let it happen is a claim in my own comment: that because ZCode's config is JSON, "did I reach every key" stops being a question. Parsing settles *structure* and says nothing about which values are secret. That comment is now corrected in place rather than quietly deleted, because it is the reason the bug exists.

Fixed in the shape mcode arrived at, not by widening the name list and hoping:

- the rewrite stays an allowlist, because it has to produce correct output;
- **origins are recognised by value, not by key name** — any `http`/`https` value moves, so `endpoint`, `host`, `url` and whatever a later ZCode invents are covered without a list;
- `accountedFor` walks the *result* and refuses anything it cannot account for: the redactor, an opaque-token shape, and any `http`-shaped value not pointing at the proxy.

On the suggestion to widen `isSecret` to mcode's `/key$/i`: that also matches `monkey`, and a *rewrite* with a false positive does not refuse, it corrupts. So the widening is case-aware — a capital or digit may precede a Titlecase or ALLCAPS word — and the value-shaped net is what carries the weight.

One thing did not transfer. mcode's opaque-token alphabet allows `-` and `_`; pointed at ZCode's own shipped `provider.example.json` it refuses the file, because `"type": "zhipu-coding-plan-api-key"` is 25 characters of exactly those and is an enum value. Every ZCode config with a coding-plan provider would have been uncaptured. The alphabet here excludes `-`, `_` and `/` — a credential has no word structure, an enum, a model id and a Windows path are short segments with separators.

Measured after the change: 17 key names × 5 credential shapes (`sk-` token, real JWT, 32 hex, AWS key id, `Bearer …`) — **zero survive**; 6 origin key names including one invented for the test — **zero stay unmoved**; the shipped example config — **carried**, enum intact, all three URLs moved.

## P1: a replay could reach a real origin the proxy cannot block

Confirmed, and worse than the built-in config route the review led with. `ZCODE_BASE_URL` is an origin, and the overlay left it alone.

Answered with a namespace sweep rather than a list, because a list is not something this adapter can honestly keep: ZCode's bundle names **200** `ZCODE_*` variables and reads them through minified constants (`Hee="ZCODE_PERSONAL_PROVIDER_CONFIG_FILE"`, read as `process.env[Hee]`), so grepping for `process.env.ZCODE_BASE_URL` finds one hit and proves nothing about the other 199. Among them: five origins, three credentials, three data directories, two proxies. `ZAI_*`, `BIGMODEL_*` and `GLM_*` go with them, and the four `OPENAI_`/`ANTHROPIC_` base-url and key variables the bundle also names are blanked by name — a prefix sweep of those would reach well past this harness.

The risk worth naming: blanking 200 variables could stop ZCode booting, the way an empty `ZCODE_BUILTIN_PROVIDER_CONFIG_FILE` does. It does not — measured, a full capture with the sweep active, 27 tools, `scrubbed clean`, artifact byte-identical to what is committed.

## Two the review did not find

**`--model` is not a flag this harness has.** `zcode --prompt … --model glm-4.6` answers `Unknown option '--model'` and prints its help. The profile passed it whenever a model was given, so `capture.mjs zcode --model glm-4.6` could not work — and that is the string the committed `regenerate` metadata contained, against `capture.mjs`'s own stated contract that it is runnable as printed. `modelFromConfig` keeps it off the command line and out of `regenerate`; passing it now exits 1 with a reason.

**The README described a capture that did not happen that way.** It said the adapter carried the operator's own origin through the proxy. The run's own generated config says otherwise — one provider, `providerId: "orca"`, pointed at the proxy. This machine has no personal provider config, so the fallback path ran and orca forwarded to its *default upstream*, which is what does not serve `glm-4.6`. Corrected, and the cwd sensitivity documented with it: ZCode's prompt embeds the working directory's git branch and `git status` output, so the capture is 9,084 characters taken in an empty repository and 9,530 taken inside this checkout.

## End to end

- full capture with the sweep active → 27 tools, `scrubbed clean`, sha256 `be796794c739c7c8`, identical to the committed artifact
- `record` `events=4 blobs=1` → `replay` `reused=1/1 exact=1 divergences=0 unmatched=0 egress=blocked`, the replayed error carrying the recording's own `requestId`
- `zcode.test.ts` 14 tests; full suite **2558 passed, 0 failed**

xizhuomengcontin:
Correction to my previous comment, and three more fixes. I pushed that round before running the path this adapter exists for, and re-reading my own claims is what turned these up rather than a new review.

## What the previous round asserted and had not run

**The carry path had never been executed by a real ZCode process.** Every check on it was `redirectedConfig` in isolation. This machine has no personal provider config, so every real run — including the one whose hash I quoted — took the `onlyOrca` fallback. So "the rewrite works" was a unit-test claim dressed as an end-to-end one.

It has a real run behind it now. Operator config pointed at a local listener, a canary key in it, `ZCODE_BASE_URL` pointed at a second listener, `ZCODE_CREDENTIAL_SECRET` set to a third canary:

```
upstream = http://127.0.0.1:46099/v1    the operator's own origin, carried through /forward/
status   = 401                          the listener answered, so the request arrived
ZCODE_BASE_URL canary                   never contacted, never recorded
both key canaries                       absent from every file in the run directory
generated apiKey                        orca-recorded
```

The recorded upstream being the operator's origin rather than orca's default is the part that matters: it is the difference between carrying a destination and substituting one.

Two false starts on the way, both worth keeping. A config with `access.type: "zhipu-coding-plan-api-key"` fails at `Model creation failed`, and so does one with `logo` as an object. Both fail identically with orca out of the picture, which is how they were ruled out as adapter faults instead of assumed to be.

## `^key$` was corrupting ZCode's own example config

I reported last round that `provider.example.json` is "carried, enum intact". True, and beside the point — it was carried **modified**. That file has `logo: { "key": "zai" }`, an image name, and the rewrite wrote `orca-recorded` over it.

This is the failure mode named in the code comment I wrote two commits ago: a rewrite with a false positive does not refuse, it corrupts. `key` and `auth` on their own are as often structure as credential, so they are out of the rewrite's name list and left to the value-shaped net; under a separator or a prefix — `api_key`, `apiKey`, `authToken` — they stay in. `auth: "api-key"` and three structural `key`s now survive a rewrite that still takes out an `apiKey` sitting beside them.

## The net compared origins with `startsWith`

`http://127.0.0.1:44100.evil.example` starts with `http://127.0.0.1:44100`. A hostname could walk through the one check standing between a real host and the trace. Parsed origins now, and a proxy URL the net cannot parse refuses rather than comparing `undefined` with `undefined` and waving everything through.

## One claim narrowed

The sweep does not remove ZCode's own defaults, it removes the operator's overrides. `resolveZCodeBaseUrl` in the wrapper is `env.ZCODE_BASE_URL?.trim() || defaultZCodeBaseUrl`, so a blank falls back to ZCode's default — which is what a normal run uses anyway, and is not where model calls go. Worth stating precisely, because the wrapper rebuilds the runtime's environment (`bin/zcode.js`, `runtimeEnvironment`) and a sweep that a wrapper overrode would have been worth nothing.

## State

`zcode.test.ts` 16 tests; full suite **2560 passed, 0 failed**. Artifact re-captured after the change, still `be796794c739c7c8`. `record` `events=4 blobs=1` then `replay` `reused=1/1 exact=1 divergences=0 unmatched=0 egress=blocked`.

xizhuomengcontin:
The P1 was real, and it reproduced against the current head rather than only the one it was written against.

The review was flagged outdated because it quoted `node.startsWith(base)`, which had already become a parsed-origin comparison in `e625db51`. That earlier change fixed a *different* hole — a hostname like `http://127.0.0.1:44100.evil.example` walking through a prefix match — and left this one untouched, so the finding was checked rather than dismissed:

```
baseUrl: https://api.example.com/v1/keys/deadbeefdeadbeefdeadbeefdeadbeef
  -> written into the run directory with the key verbatim
```

Both nets looked away for the same reason the review gives. The shape check never ran, because a forwarded value's origin *is* the proxy's and the walk returned early. And 32 hex characters is the one shape the redactor is documented in this very file to miss, its maximum Shannon entropy being exactly the threshold. So the exemption was granted by the act of forwarding, which is the opposite of what forwarding is supposed to prove.

`accountedFor` now decodes the `/forward/` target and runs the shape check on the origin the value replaced. Refused after the change, at two path depths.

What had to keep working, and does: an origin whose path carries nothing, and a value already pointing at the proxy. Three shapes that already refused still refuse by their previous routes — `sk-` and a 40-character run via the redactor, a query string because `/forward/` declines to carry one.

Carried path re-measured end to end after the change, operator config pointed at a local listener with three canaries:

```
upstream = http://127.0.0.1:46099/v1    the operator's origin, carried
status   = 401                          the listener answered
gateway received                        1 POST
ZCODE_BASE_URL canary                   never contacted
both key canaries                       0 hits anywhere in the run directory
```

`zcode.test.ts` 17 tests; full suite **2561 passed, 0 failed**; artifact re-captured and still `be796794c739c7c8`.

No further pushes queued, so the review on `02430c33` should not go stale mid-run.

PR #112, opened 2026-09-21
fix(adapters): a separator is not a net, and a capture is not a fact about the machine

xizhuomengcontin:
## Orca-Code-Review — push 3

|---|---|---|
| P0 | 0 | 0 |
| P1 | 0 | 0 |
| P2 | 0 | 0 |
| P3 | 0 | 0 |

✅ no blocking findings

## What this changes

Two follow-ups to #111. The first is a §7 hole that a review found and the merge did not wait for; the second is a `regenerate` command that corrupted the artifact it was supposed to reproduce. Both were found by running the paths #111 claimed rather than the ones it happened to exercise.

### A credential broken by separators walked through both nets

`OPAQUE_TOKEN` excluded `-` and `_`. That exclusion was paid for: ZCode's own shipped `provider.example.json` has `"type": "zhipu-coding-plan-api-key"`, twenty-five characters of exactly those, and removing the separators was how that file stopped being refused. The comment justifying it claimed the redactor was the net for anything written with separators instead.

Measured, it is not:

| value | length | redactor | why it declines |
|---|---|---|---|
| `abc123-abc123-abc123-abc123` | 27 | miss | entropy ~2.8 bits/char against a 4.0 threshold |
| `deadbeef-deadbeef-deadbeef` | 26 | miss | no digit, so `looksRandom` is false |
| `k1a2b3c4_k1a2b3c4_k1a2b3c4` | 26 | miss | same |
| `AbCd-EfGh-IjKl-MnOp-QrSt-UvWx` | 29 | miss | same |

**And it is reachable through a documented field, not a contrived one.** `api.headers` is described in ZCode's own `PROVIDER_CONFIG.md` as an arbitrary string map — *"Additional HTTP headers, e.g. `{"X-Client-Name":"zcode-cli"}`"* — which is where a gateway's signed header goes. In a real recording, before the fix:

```
X-Gateway-Sig:       abc123-abc123-abc123-abc123   -> run directory, 1 hit
X-Request-Signature: deadbeef-deadbeef-deadbeef    -> run directory, 1 hit
X-Tenant-Id:         k1a2b3c4_k1a2b3c4_k1a2b3c4    -> run directory, 1 hit
access.apiKey:       sk-live-PLAIN-CANARY-1        -> 0 hits (the rewrite was fine all along)
```

`capture.mjs` copies the run directory into `capture/<model>/trace/`, which this repo documents as unscrubbed and `orca push` refuses to ship. The `apiKey` being handled correctly the whole time is what made the rest look fine.

The separators are back, and the false positive is answered where it belongs. `STRUCTURAL_FIELD` names the four fields in ZCode's schema whose values are identifiers — `type`, `modelId`, `personalModelIds`, `modelOrder` — exempt from the shape check only, and still read by the redactor. Each is there because it was measured: `type` carries a 25-character enum, and a provider-qualified model id is 26 characters after the `/` that ends the run. **A wrong entry in that list is bounded to one field name; a missing character class was not.**

### A capture recorded the directory it was taken in

ZCode's prompt embeds three things about the working directory rather than the harness: the git branch, the `git status` output, and the name of its per-project memory directory, which is `<basename>-<16 hex of the absolute path>`.

So the committed artifact was a fact about one directory on one machine — and running the `regenerate` command as printed **overwrote it** with whatever the operator's tree looked like:

```
node capture/capture.mjs zcode        # the string capture.mjs itself printed
  -> 9,501 chars, mirrored over prompt/ZCODE/…, be796794… -> 1816b8fd…
```

`capture.mjs` already holds the principle this violates, in the `{{RUN_ID}}` rule: *"A run id makes the artifact different on every capture, which would make `regenerate` produce a diff every time and mean nothing when it did not."* So all three are scrubbed the same way, which is stronger than capturing from a fixed disposable directory would have been — a fixed directory still bakes in its own path hash, and a random one changes the artifact on every run.

After the change:

```
in this repository, dirty tree, branch fix/zcode-separator-net  -> ec84e8e90cc70992
in an empty repository on master                                 -> ec84e8e90cc70992
three runs across two directories                                -> one hash
regenerate against the committed artifact                        -> 0 lines of diff
```

The status rule had to learn where a status block ends. Claude Code's prompt puts `Recent commits:` directly after it with no blank line between, and the first version — every non-empty line — swallowed that heading and the commits under it. Caught against the committed Crush capture, which has that shape; each line now has to look like an entry, and `Recent commits:` does not, because `R` is a status letter but `e` is not a space.

### Four already-committed artifacts had the same thing

Adding the rules is what surfaced them. None is a credential; all of it is the capture machine's working state in a published file:

| artifact | branch | dirty paths |
|---|---|---|
| `CLAUDECODE/claude-opus-5` | `capture-opencode` | 15, including `packages/cli/src/tls-capture.ts` |
| `CRUSH/deepseek-v4.1-flash` | `prompts/openrouter-top` | 6, including `?? .venv/` |
| `CLAUDECODE/claude-fable-5-1` | `capture-system-prompts` | 3 |
| `CLAUDECODE/claude-opus-4-8` | `main` | 2 |

Re-scrubbed by applying the new rules to the text rather than re-capturing, and the Crush capture's local source with it, so `--index` does not put it back. The as-sent figures in the table are unchanged, since scrubbing happens after the prompt is sent.

## Tests

`packages/adapters/test/zcode.test.ts`, 19 tests, three new:

- **`refuses a signed header, which is where a real one lives`** — the reachable case above, as a unit test, so the thing that actually happened cannot happen again quietly.
- **`lets an identifier be as long as identifiers get`** — the cost of putting the separators back, pinned: a 26-character model id is carried under `modelId` and refused under a name nothing accounts for.
- the four separator shapes added to the existing refusal matrix, which previously used only shapes with no separators — which is why it passed.

Measured beyond the suite:

- 5 separator shapes × 5 field names through `redirectedConfig`: none survive
- the `api.headers` recording re-run for real: all four canaries absent from the run directory, the config falling back to `providerId: "orca"` at the proxy, and the run still captured
- `provider.example.json` still carried, enum intact, all three URLs moved
- the two holes #111 closed are still closed: a 32-hex key inside a forwarded path, and a hostname that starts with the proxy's own URL
- `record` `events=4 blobs=1`, then `replay` `reused=1/1 exact=1 divergences=0 unmatched=0 egress=blocked`
- artifact leak scan across 11 patterns: 0 hits

## Checklist

- [x] `npm run check` passes locally — `typecheck` and `vitest` clean (**2563 passed, 0 failed**). `fmt:check` is green on every tracked file; it reports untracked local `capture/<run>/` artifacts, which are gitignored and never reach CI.
- [x] Tests were written before the implementation — for the two named above, yes: the `api.headers` case was written from the real recording that leaked, before the alphabet changed.
- [x] Spec and schema updated together, if the trace format changed — it did not.
- [x] Commits signed off


xizhuomengcontin:
/orcacode-review

Asking for a second read of `0d175392` rather than treating the first as settled. Earlier on #111 two reviews of the same commit disagreed — 07:15 reported a P1 and 07:43 reported no findings, with no push between them — so one clean run is weaker evidence than it looks. This one is merged either way; a finding here is still worth a follow-up.

PR #114, opened 2026-09-21
fix(cli): name the listing's fields after a response, not after a guess

xizhuomengcontin:
## Orca-Code-Review — push 1

| Severity | Count |
|---|---|
| P0 | 0 |
| P1 | 0 |
| P2 | 0 |
| P3 | 0 |

✅ no blocking findings

Stacked on **#113** — that PR's `orca list --remote` has a column that is empty on every row, and this is the correction. Base is `feat/list-remote`, so the diff here is only the fix; it will retarget to `main` if #113 merges first.

## The defect

`STARTED` was mapped from `created_at`. **No captured gateway response contains that field.** A listing item is:

```json
{ "run_key": "run_67a7ce30bd6a", "source": "upload", "client_app": "claude-code",
  "turns": 5, "tool_calls": 3, "models": ["claude-opus-5"],
  "layers": ["model", "fs"], "outcome": "exit 0", "bytes": 201620 }
```

No timestamp at all. The *detail* endpoint's `session` object has `first_ts` / `last_ts`; the listing item has neither. So the column took the missing-value path on every row, for every run — and before the em-dash guard landed in the same PR it would have read `1970-01-01 00:00` instead.

Two other things came from the same guess:

- `outcome` is the run's **exit status as a string** — `exit 0`, the same convention `orca show` already prints. The tests and README example used `ok` / `error`, which the gateway never sends.
- The doc comment claimed the console's columns "minus the two a terminal cannot use" while actually dropping **four**. The console shows Run, Started, Source, App / agent, Turns, **Tools**, **Layers**, Models, Outcome, Cost, Open.

## Why the tests did not catch it

The fixture was written from the mapping rather than from a response. It carried `created_at` and an outcome of `ok`, so all twelve tests confirmed that the code agreed with the person who wrote both sides of it. That is the actual bug; the empty column is its symptom.

The fixture is now a captured item, field for field, and it is the thing every test in the file builds on.

## The fix

**`startedAt` reads whichever of `started_at`, `created_at`, `first_ts` a deployment sends**, and renders `—` when there is none. Not a swap of one guess for another: the console *has* a Started column, so a deployment may well fill one, and which name it uses is not something this should have to be right about in advance. One `it.each` covers all three.

**LAYERS and TOOLS are now shown** — both already in the response, both in the console:

```console
$ orca list --remote --limit 2
RUN                           STARTED           SOURCE   APP          TURNS  TOOLS  LAYERS       MODELS                    OUTCOME
run_67a7ce30bd6a              2026-09-15 08:03  upload   claude-code  5      3      model,fs     claude-opus-5             exit 0
run_c3b56f9910ee6050c468aa7c  2026-09-14 07:03  gateway  —            12     0      model,route  claude-sonnet-4-5, gpt-5  exit 1
```

LAYERS is the one that matters. The old comment justified SOURCE as a stand-in for which layers a run holds — it is not one. `model` and `route` come from a gateway recording; `fs`, `shell`, `mcp` and `net` from a run recorded here. It is what tells you whether a run is worth the download, which is the entire point of listing before pulling. SOURCE stays beside it because where a run came from is a different question from what it holds.

Cost and Open stay out: one belongs to billing, the other is a link. `bytes` also stays out — nine columns is already 131 characters, in line with what local `orca list` prints with a full `DIR` path.

The `—` for an unreported field is kept rather than hiding the column, because it is already this table's marker for "the gateway did not report this" in three other places, and a stable header is worth more than nine characters.

## Verification

16 tests (12 → 16), every one now built on the captured item. Revert-goes-red checked for each new behaviour: the tolerant timestamp read (2 red), the two new columns (3 red).

End to end through the built binary against **both** shapes — a host serving the captured item verbatim, and one that adds `started_at`:

```
no timestamp     STARTED reads —, every other column filled
with started_at  STARTED reads 2026-09-15 08:03
401              "Invalid API key" — the gateway's own words, verified against the real host
--source         forwarded, filtered by the gateway
```

Gates: `prettier --check .`, `tsc --build --force`, `scripts/conformance.mjs`, `scripts/fidelity.mjs --check` — all green. Full suite **2714 passed, 0 failed**.

## One thing still unverified

Whether `www.orcarouter.ai`'s current listing includes a timestamp under *any* name. The captured response is from a deployment that sends none, and I have no `replay`-scoped key in hand to re-probe. The fix is correct either way — that is why it reads three names and degrades to `—` — but if it does send one under a fourth name, the column will still be empty and I would want to know.


xizhuomengcontin:
**Correction to this PR's central claim.**

This PR says *"No captured gateway response carries `created_at`"* and that the STARTED column *"took the missing-value path on every row, for every run"*. That was true of the deployment the capture came from — an older one that sends nine fields and no timestamp. **It is false of production.** Read through `www.orcarouter.ai`'s own console on 2026-09-23, the listing sends `created_at`, `first_ts`, `last_ts`, `updated_at` and `expire_at`, all in seconds. The code as it was in #113, reading `created_at`, would have shown a date on production.

What this PR changed still holds up — reading several field names defensively is the reason the code was never wrong on either deployment. The claim about why it was needed was overstated, and I wrote a claim about every deployment from a capture of one. That is the mistake this PR diagnoses in a fixture, one level up.

Two further statements here are also wrong for production gateway recordings: `outcome` is `end_turn` or `error` rather than an exit status (only pushed runs report `exit 0`), and `client_app` is `Unknown`, `curl` or `Node` rather than empty.

Production also showed that `created_at` is the wrong field to prefer for STARTED: it is when the gateway stored the row, which for a pushed run is the push, not the run's start. All of this is corrected in **#120**, with a production-shaped fixture and the README example regenerated from the CLI's real output.

